PR #3, opened 2016-11-19
restructure files, adds a few comments/docs

ZacLN:
@davidanthoff I've moved everything around a bit. All code specific to providers is moved to the relevant self contained files, all document handling functions are grouped together and I've moved all things to do with the LanguageServerInstance together

Let me know your thoughts!


davidanthoff:
I'm not sure. I think it was good that all the types that correspond to the actual protocol were in one file, so that it is clear that those types are the ones that get serialized to JSON and that one can't just edit them and e.g. add fields, or rename them etc. So that file essentially would track the definitions in the LSP definition.

I would actually also prefer that for example the type and methods that handle the internal representation of documents in the LS don't depend on the types from the protocol definition. I think that kind of layering makes things more future proof and also makes it a lot easier to write effective unit tests.

ZacLN:
Hmm, that seems to make sense to me as well. 

For hover methods for example, do you mean that we should have a `parse_params` method that simply spits out the linear index of the `Char` and the `uri::String` -> we do get our results and then **all** that the `process` function does is wrap our returned results into the expected structure and convert to JSON?

davidanthoff:
Yeah, I guess. I think one way to think about layering here is: if another editor came about that wanted to use the functionality here, but not use the language server protocol, but say some other protocol, would that work? I.e. ideally most of the real functionality would not depend on the protocol implementation.

davidanthoff:
@ZacLN We can probably close this?

PR #4, opened 2016-11-25
Make Document type more self-contained

davidanthoff:
Not ready for merging, I'm in the middle of this. Just wanted to give @ZacLN a heads up what I'm working on right now.

I'll write up an explanation about these changes once they are done. The broad idea is to a) encapsulate the ``Document`` type a bit more, so that it a) can be unit tested more easily, b) cut down on the number of allocations and c) make more use of the julia ``String`` code. But I'll provide a more thorough write up later. I hope to get this done later today.

### DIFF
--- a/src/document.jl
+++ b/src/document.jl
@@ -17,7 +17,7 @@ function get_line(doc::Document, line::Int)
 
     if length(line_offsets)>0
         start_offset = line_offsets[line]
-        if length(line_offsets)>line
+        if length(line_offsets)>line            
             end_offset = line_offsets[line+1]-1
         else
             end_offset = endof(doc._content)
@@ -69,7 +69,7 @@ function get_line_offsets(doc::Document)
 		end
 
 
-        if is_line_start && text.length > 0
+        if is_line_start && length(text) > 0
 		    push!(line_offsets, endof(text))
 		end
 
--- a/src/document.jl
+++ b/src/document.jl
@@ -17,7 +17,7 @@ function get_line(doc::Document, line::Int)
 
     if length(line_offsets)>0
         start_offset = line_offsets[line]
-        if length(line_offsets)>line            
+        if length(line_offsets)>line
             end_offset = line_offsets[line+1]-1
         else
             end_offset = endof(doc._content)
@@ -44,7 +44,7 @@ function update(doc::Document, start_line::Int, start_character::Int, length::In
     for i=1:length
         end_offset = nextind(text,end_offset)
     end
-    
+
     doc._content = string(doc._content[1:start_offset-1], new_text, doc._content[end_offset:end])
     doc._line_offsets = Nullable{Vector{Int}}()
 end
@@ -85,5 +85,5 @@ function get_position_at(doc::Document, offset::Int)
             return line-1,offset-line_offsets[line-1] + 1
         end
     end
-	return length(line_offsets), offset - line_offsets[end] + 1
+    return length(line_offsets), offset - line_offsets[end] + 1
 end
--- a/src/document.jl
+++ b/src/document.jl
@@ -42,6 +42,7 @@ function update(doc::Document, start_line::Int, start_character::Int, end_line::
     end_offset = get_offset(doc, end_line, end_character) 
     
     doc._content = string(doc._content[1:start_offset], new_text, doc._content[end_offset+1:end])
+    doc._line_offsets = Nullable{Vector{Int}}()
 end
 
 function _get_line_offsets(doc::Document)
--- a/src/document.jl
+++ b/src/document.jl
@@ -1,4 +1,75 @@
 type Document
-    data::Vector{UInt8}
+    _content::String
+    _line_offsets::Nullable{Vector{Int}}
     blocks::Vector{Any}
+
+    function Document(text::AbstractString)
+        return new(text, Nullable{Vector{Int}}(), [])
+    end
+end
+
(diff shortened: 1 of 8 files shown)


ZacLN:
Cool, this looks cleaner

davidanthoff:
@ZacLN This is ready for review now.

Here are the major changes:
- I'm back to using ``String`` as the format to store a doc. I don't see any benefit of using an ``Vector{UInt8}``, really. ``String`` is just a wrapper around a ``Vector{UInt8}``, so there is no real difference in the storage format, but we lose all the convenience of using ``String`` as the type. Also, a major benefit of ``String`` is that ``s[x:y]`` will return a ``SubString`` that doesn't allocate a new string but is essentially a view. This should be especially beneficial for things like ``get_line``: it should now no longer allocate a new string.
- The old incremental doc update allocated quite a bit more memory than necessary because it allocated lots of intermediate objects. I think the current implementation should be as good as we can get it, unless we go to some way more fancy data structure.
- The general idea is that all access to the ``Document`` members should go via the ``methods`` defined in document.jl, i.e. ideally no more direct access to the fields from e.g. parse code etc. That should make it easier to later exchange the data structure that we use, if we want to change it.
- I cached the line break computation, in the hopes of speeding things up a bit.
- Things are unit tested :)
- I changed things to be 1-based in the julia part of the code... I hope that doesn't bite as later, but I kind of hope we can keep the 0-based stuff in the language server protocol layer, and have the code that does the real work follow julia conventions and use 1-based indexing. I'm not a 100% I got this consistently right...

ZacLN:
I'm very happy with all that! We really do need more consistency in 0/1 based and  line-char/linear indexing as it's doing my head in and I've just noticed that half the code is still miss-handling unicode characters.

- 'parse' stuff should all work on TokenStreams from now so will be unaffected. 
- The location info for expressions/blocks is measured in UInt8 offsets which sadly adds a third measure.

Is this far off being mergable?

ZacLN:
I've made a big mistake, this ain't ready yet

src/document.jl
```diff
@@ -1,4 +1,89 @@
...
+        else
+            end_offset = endof(doc._content)
+        end
+        return doc._content[start_offset:end_offset]
+    else
+        return ""
+    end
+end
+
+function get_offset(doc::Document, line::Int, character::Int)
```
ZacLN: On an empty document this attempts to access `line_offsets` at 0, maybe needs to be rejigged to include the 0 offset for line 1.

src/document.jl
```diff
@@ -1,4 +1,89 @@
...
+end
+
+function get_position_at(doc::Document, offset::Int)
+    line_offsets = get_line_offsets(doc)
+    for (line, line_offset) in enumerate(line_offsets)
+        if offset<line_offset
+            return line-1,offset-line_offsets[line-1] + 1
+        end
+    end
+    return length(line_offsets), offset - line_offsets[end] + 1
```
ZacLN: Again, attempts to access an array that can be empty

src/document.jl
```diff
@@ -1,4 +1,89 @@
...
+
+function update(doc::Document, start_line::Int, start_character::Int, length::Int, new_text::AbstractString)
+    text = doc._content
+    start_offset = get_offset(doc, start_line, start_character)
+    end_offset = start_offset
+    for i=1:length
+        end_offset = nextind(text,end_offset)
+    end
+
+    doc._content = string(doc._content[1:start_offset-1], new_text, doc._content[end_offset:end])
```
ZacLN: Accesses possibly empty string using `end`.

src/document.jl
davidanthoff: I'm tempted to leave this as is and treat any call to this function as an error if the doc is empty... I kind of feel this should be handled by callers of this function, i.e. they should never call ``get_offset`` if the doc is empty.

src/document.jl
davidanthoff: I think ``end`` should be fine here. For example ``x[1:end]`` works for an empty array.

src/document.jl
davidanthoff: I think here also, this function should just not be called if the doc is empty, so I think this needs to be handled in the caller.

davidanthoff:
I did find a bunch of errors related to empty docs, though. Fixing those bugs now.

davidanthoff:
> The location info for expressions/blocks is measured in UInt8 offsets which sadly adds a third measure.

But that is fine, that is how one actually indexes into the ``String`` returned from ``get_text``, right? I think essentially we have: 1) 1-based line/character combos, 2) 1-based byte-offsets and 3) 0-based line/character combos. We should try to keep 3) as much as possible in the protocol implementation layer, and use 1) and 2) mostly internally, right?

src/document.jl
ZacLN: I think that if you were to either 1) type into an empty document and then delete what you had typed or 2)  demand a completion (ctrl+space on my setup) on an empty document then a completion provider would be requested for an empty document

ZacLN:
Yep that's all sensible, I'd say we should try and keep all references to the line/character 0 offset stuff inside `process` functions

src/document.jl
davidanthoff: I think #13 should handle both of these cases correctly. Essentially the completion provider now doesn't even call ``get_offset`` if the cursor is on an empty line.

PR #7, opened 2016-11-26
fix latex completions

ZacLN:


src/provider_completions.jl
```diff
@@ -1,7 +1,23 @@
 function process(r::JSONRPC.Request{Val{Symbol("textDocument/completion")},TextDocumentPositionParams}, server)
     tdpp = r.params
     line = get_line(tdpp, server)
-    comp = Base.REPLCompletions.completions(line, chr2ind(line,tdpp.position.character))[1]
+    io = IOBuffer(reverse(line[1:chr2ind(line,tdpp.position.character)]))
```
davidanthoff: Why use an ``IOBuffer`` here? Why not just iterate over the characters of the string via ``for c in s`` instead of the ``while`` loop later?

src/provider_completions.jl
```diff
@@ -1,7 +1,23 @@
...
+    word = Char[]
```
davidanthoff: Actually, one should use an ``IOBuffer`` when constructing a ``String`` in many steps. So I would use an ``IOBuffer`` here, then convert that to a ``String`` using ``takebuf_string`` and then reverse the ``String``.

That should be much more efficient than using ``unshift!``.

src/provider_completions.jl
```diff
@@ -1,7 +1,23 @@
...
+            unshift!(word, c)
+            break
+        end
+        if !(Base.is_id_char(c) || c=='.' || c=='_' || c=='^')
+            break
+        end
+        unshift!(word, c)
+    end
+    
+    str = String(word)
```
davidanthoff: Replace this with ``reverse(takebuf_string(word))`` once you use ``IOBuffer``.

src/provider_completions.jl
```diff
@@ -1,7 +1,23 @@
...
+    prefix = str[1:findlast(word,'.')]
+    comp = Base.REPLCompletions.completions(str,length(str))[1]
```
davidanthoff: ``length`` is incorrect, because you need to pass the byte count here, not the character count, I believe.

davidanthoff:
@ZacLN This now has merge conflicts, can you resolve them? Thanks!

src/provider_completions.jl
```diff
@@ -1,38 +1,56 @@
...
         end
+        if !(Base.is_id_char(c) || c=='.' || c=='_' || c=='^')
+            break
+        end
+        write(word, c)
+    end
+    
+    str = reverse(takebuf_string(word))
+    prefix = str[1:findlast(word,'.')]
+    comp = Base.REPLCompletions.completions(str,length(str.data))[1]
```
davidanthoff: Maybe use ``endof(str)`` instead of ``length`` here? Probably safer to not assume a certain storage format here, I wouldn't be surprised if they changed the internals of ``Strings`` again at some future point.

PR #8, opened 2016-11-27
WIP: use JuliaParser, get full namespace context

ZacLN:
WIP: not ready to merge

This switches over to using a forked JuliaParser rather than Base.parse. (Requires [this PR](https://github.com/JuliaEditorSupport/julia-vscode/pull/109) on julia-vscode to be merged).

- Code is now simply stored as the parsed AST of the document with string-> expression location relations stored in the `.typ` field of any given sub-expression within the AST.
- Currently updates the AST in the same way as the previous iteration but the structure is in place to update (re-parse) only the most minimal sub-expressions that are changed by a `textDocument\didChange` event.
- User functionality is slightly regressed from the previous position, hovers just show the position and definition of a local variable and field indexing hasn't been re-implemented.

@davidanthoff (or others), if you could check that this works as is, that would be great. I'm going to try and re-implement all the old functionality today then will add: 
- completions/signatures/goto def for user defined variables
- optimize re-parsing
- include AST of `include(...)` statements
- return `Diagnostic`s on parsing errors

I think this should probably be merged after #4 and any other structural changes we want to make?


davidanthoff:
I just tried it and it seems to work. This is really cool! Do you think there is a way to merge your changes to JuliaParser into the main repo? It would be a lot easier if we don't have to maintain our own fork... If you open a PR, do ping me on it.

Oh, and we should start to write unit tests for code  going forward :)

ZacLN:
I think the JuliaParser stuff is unlikely to get in anytime soon, it adds too much mess to the output. 

We're level with/beyond where we were with features now on this so I'm going to add completions for local (file defined) variables and add some tests then it'll be ready to go.

We need to have a think about how we want to handle workspace wide parsing and user included Modules, I'm a bit wary of just parsing everything and including it across all files. I think we need some set of rules that determine whether some arbitrary file is related to the one currently open. Maybe based on `include(..)`'s across the project?

davidanthoff:
This looks very cool. I've added two new issues that discuss some of the more thorny aspects of this (#18 and #19), but my vote would be that we go ahead with this PR before we resolve these, given that it seems to improve the situation just for single files already. And for now we can live with the forked version of JuliaParser and then try to find a solution later, right?

davidanthoff:
Ah, another thing: to get the CI builds working, I would suggest you install the forked JuliaParser in the travis and appveyor files. I did something similar [here](https://github.com/davidanthoff/Query.jl/blob/master/.travis.yml#L12) and [here](https://github.com/davidanthoff/Query.jl/blob/master/appveyor.yml#L28). You would have to add another command to checkout the correct branch, but I think that is the simplest way to get the CI testing working.

ZacLN:
Yep I think merge and come back to the issue later. Thanks for the tips on on getting it to pass tests! though it's now failing in incomprehensible ways...

davidanthoff:
The appveyor build fails for the following reason: the constructor for ``Position`` requires ``line`` and ``character`` to be of type ``Int``. On win32 this ends up being ``Int32``. Apparently the JSON package returns ``Int64`` for numbers, irrespective of the platform, so the ``Position`` constructor that takes a ``Dict`` will try to call the inner constructor with two ``Int64`` arguments. I think at the end of the day this is an error in JSON, it should use the platform appropriate integer type.

As a workaround, for now, I would convert the types in the constructor that takes the ``Dict``. I'll open a PR against JSON.jl in the meantime.

davidanthoff:
See JuliaIO/JSON.jl#178.

davidanthoff:
Ok, the JSON behavior actually makes sense. So we should just change the arguments of the ``Position`` constructor to ``Integer``, that should also fix the win32 build.

src/protocol.jl
```diff
@@ -1,9 +1,9 @@
 # Position
 type Position
-    line::Int
-    character::Int
+    line::Integer
+    character::Integer
```
davidanthoff: We should use concrete leave types here, i.e. in type declarations, otherwise these values will always be boxed, which is inefficient. I think ``Integer`` is the correct type in the message signature (so that we dispatch correctly), but not in type declarations. I would go with ``Int`` here, that should be faster on x86, and should be plenty in terms of domain.

ZacLN:
This is ready to go! This is stable and working on my machine now and my inability to get a deprecation fixing/non behaviour changing PR into JuliaParser makes me think it's not worth waiting on a resolution to #18 for now. 

As it stands this:

- Uses a forked `JuliaParser.jl` adding more detailed code location information to `Expr`s.
- changes how a parsed document is stored (from `Vector` -> `Expr`)
- can construct the scope of file at any given point 
- Improves hovers/completions/goto for file defined variables
- parses files defining modules properly
- fixes a latex completions error
- removes the `get_docs` function 
- adds tests

There's a further PR that implements caching for `Base`/`Core` which substantially changes/simplifies how information for hovers/signatures/etc... are handled but I don't really want this to grow any further

davidanthoff:
Cool! I just tried it and I seem to get this error
````
ERROR: LoadError: MethodError: no method matching start(::Type{Any})
Closest candidates are:
  start(!Matched::SimpleVector) at essentials.jl:170
  start(!Matched::Base.MethodList) at reflection.jl:258
  start(!Matched::IntSet) at intset.jl:184
  ...
 in first(::Type{T}) at .\abstractarray.jl:132
 in getsyminfo(::Expr, ::Array{LanguageServer.SymbolInformation,1}, ::String, ::LanguageServer.Document, ::String) at C:\Users\anthoff\.julia\v0.5\LanguageServer\src\provider_symbols.jl:33
 in process(::LanguageServer.JSONRPC.Request{Val{Symbol("textDocument/documentSymbol")},LanguageServer.DocumentSymbolParams}, ::LanguageServer.LanguageServerInstance) at C:\Users\anthoff\.julia\v0.5\LanguageServer\src\provider_symbols.jl:12
 in run(::LanguageServer.LanguageServerInstance) at C:\Users\anthoff\.julia\v0.5\LanguageServer\src\languageserverinstance.jl:28
 in include_from_node1(::String) at .\loading.jl:488
 in process_options(::Base.JLOptions) at .\client.jl:262
 in _start() at .\client.jl:318
while loading c:\Users\anthoff\source\julia-vscode\scripts\languageserver\main.jl, in expression starting on line 29
````
a lot when I try to view the local symbols. Not in all files, but a fair bit, for example in [this](https://github.com/davidanthoff/fund.jl/blob/master/src/fund.jl) file.

I'll give this more of try over today.

ZacLN:
Ah good catch! let me know what else you find

davidanthoff:
Another thing (that is probably there in the released version already): hover detection seems one character off. If I have the mouse over the first character of an identifier I don't get the hover info for that, but if I have the mouse over the next character after an identifier I still get the hover help for that identifier. I bet this is a 0 vs 1 based indexing issue.

ZacLN:
fantastic. I'll add some notes to the code and try to clean it up a bit but basically:
**`staticanalysis.jl`**
`parseblocks`: does the partial parsing as before, its a bit messy and can be improved
`get_type`\`get_field`: allows you to know if a symbol represents a struct and what types they are.

**`namespace.jl`**
`get_names{t}(ex, ...)`: takes an expression, checks whether it defines any variable and if you are currently located inside it (`loc` represents your position in the code) starts grabbing any locally defined variables.

Currently it misses quite a few cases but it's easy to improve them.

PR #12, opened 2016-11-28
Revert "Make Document type more self-contained"

ZacLN:
My merge was a mistake, this is not ready
Reverts JuliaEditorSupport/LanguageServer.jl#4


davidanthoff:
Ah, I don't doing a revert is a good idea when we actually want to merge this eventually, it makes things tricky with git... :) I'm actually tempted to force push ``master`` back to 9e1281ba64bdbd27e29ad, if that is ok with you? And then, in the future, we should never ever push anything to ``master`` directly, even if it is just a revert like this.

davidanthoff:
If we do merge something into master that we realize is not done, we can just fix it on master, but reverting should really only happen if we don't want to see that stuff ever on master.

davidanthoff:
@ZacLN  Alright, I've force pushed master. So I would suggest you delete your local ``master`` branch, then fetch from origin, then checkout ``master`` again.

ZacLN:
Ah ok, I may have panicked a bit in the early hours of the morning. Did my notes make sense?

davidanthoff:
Yes, they did, I found a whole bunch of bugs. I'll have fixes in < 30 secs ;)

ZacLN:
Ha! very good

PR #13, opened 2016-11-28
Make Document type more self-contained 2

davidanthoff:
Second attempt at #4. This should fix the bugs related to empty documents, at least the ones I found. Ready for review, @ZacLN.

### DIFF
--- a/src/document.jl
+++ b/src/document.jl
@@ -17,7 +17,7 @@ function get_line(doc::Document, line::Int)
 
     if length(line_offsets)>0
         start_offset = line_offsets[line]
-        if length(line_offsets)>line
+        if length(line_offsets)>line            
             end_offset = line_offsets[line+1]-1
         else
             end_offset = endof(doc._content)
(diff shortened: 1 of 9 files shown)


src/document.jl
```diff
@@ -1,4 +1,89 @@
...
+        end
+        return doc._content[start_offset:end_offset]
+    else
+        return ""
+    end
+end
+
+function get_offset(doc::Document, line::Int, character::Int)
+    line_offsets = get_line_offsets(doc)
+    current_offset = line_offsets[line]
```
ZacLN: I'm hitting an out of bounds error following:
1. new document
2. new line
3. any character

davidanthoff: Ha, that uncovered a bug in ``get_line_offsets``. That one actually now returns a line offset that might not be a valid index into the content ``String``, namely if there is a trailing new line. I hope I fixed this with the latest push here.

PR #16, opened 2016-11-30
add unicode symbol as doc, cleanup code

ZacLN:
Including the unicode character as the label doesn't seem to work so I've added it as a doc.


davidanthoff:
Perfect!

PR #17, opened 2016-12-01
Add toggle option for linter

ZacLN:


src/provider_misc.jl
```diff
@@ -69,10 +69,32 @@ end
...
+function process(r::JSONRPC.Request{Val{Symbol("\$/setTraceNotification")},Dict{String,Any}}, server)
+end
+
+function JSONRPC.parse_params(::Type{Val{Symbol("\$/setTraceNotification")}}, params)
+    return Any(params)
+end
+
+function process(r::JSONRPC.Request{Val{Symbol("workspace/didChangeConfiguration")},Dict{String,Any}}, server)
+    if isempty(r.params["settings"])
+        server.runlinter=false
```
davidanthoff: We could send one ``PublishDiagnostics`` notification for each file here that has an empty ``diagnostics`` field, and that would probably remove all existing lint messages, right?

ZacLN: Yep! That'll wokr

PR #22, opened 2016-12-07
Workspace level doc handling

davidanthoff:
Goes together with JuliaEditorSupport/julia-vscode#116.

### DIFF
--- a/src/provider_misc.jl
+++ b/src/provider_misc.jl
@@ -34,7 +34,7 @@ end
 
 function process(r::JSONRPC.Request{Val{Symbol("textDocument/didOpen")},DidOpenTextDocumentParams}, server)
     uri = r.params.textDocument.uri
-    if !haskey(server.documents, uri)    
+    if !haskey(server.documents, uri)
         server.documents[uri] = Document(uri, r.params.textDocument.text, false)
     end
     doc = server.documents[uri]
@@ -91,11 +91,11 @@ function process(r::JSONRPC.Request{Val{Symbol("workspace/didChangeWatchedFiles"
         if change._type==FileChangeType_Created || change._type==FileChangeType_Changed
             filepath = uri2filepath(uri)
             content = String(read(filepath))
-            server.documents[uri] = Document(uri, content, true)     
+            server.documents[uri] = Document(uri, content, true)
 
-            if should_file_be_linted(uri, server) 
-                process_diagnostics(uri, server) 
-            end                   
+            if should_file_be_linted(uri, server)
+                process_diagnostics(uri, server)
+            end
         elseif change._type==FileChangeType_Deleted
             delete!(server.documents, uri)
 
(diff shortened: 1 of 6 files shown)


ZacLN:
How far off are we with this? I'm using part of this to enable the first stage of the handling of `include` calls

davidanthoff:
I think I need to find maybe an hour or two to finish this. Hopefully this week, a bit tricky right now, end of semester etc.

ZacLN:
Yep yep fair enough!

davidanthoff:
@ZacLN Ready for review!

src/provider_misc.jl
```diff
@@ -12,6 +12,18 @@ const serverCapabilities = ServerCapabilities(
 
 function process(r::JSONRPC.Request{Val{Symbol("initialize")},Dict{String,Any}}, server)
     server.rootPath=haskey(r.params,"rootPath") ? r.params["rootPath"] : ""
+    if server.rootPath!=""
+        for (root, dirs, files) in walkdir(server.rootPath)
+            for file in files
+                if splitext(file)[2]==".jl"
+                    filepath = joinpath(root, file)
+                    uri = string("file:///", replace(replace(filepath, '\\', '/'), ":", "%3A"))
```
ZacLN: linux (and MacOS?) already use forward slashes so we need something along the lines of: 
`uri = is_windows() ? string("file:///", replace(replace(filepath, '\\', '/'), ":", "%3A")) : string("file://", filepath)
            println(uri)`

src/provider_misc.jl
```diff
@@ -21,7 +33,13 @@ function JSONRPC.parse_params(::Type{Val{Symbol("initialize")}}, params)
...
 function process(r::JSONRPC.Request{Val{Symbol("textDocument/didOpen")},DidOpenTextDocumentParams}, server)
-    server.documents[r.params.textDocument.uri] = Document(r.params.textDocument.text)
+    uri = r.params.textDocument.uri
+    if !haskey(server.documents, uri)
+        server.documents[uri] = Document(uri, r.params.textDocument.text, false)
+    end
+    doc = server.documents[uri]
+    set_open_in_editor(doc, true)
+
     parseblocks(r.params.textDocument.uri, server)
```
ZacLN: use `uri`?

davidanthoff:
@ZacLN Thanks for the review, I think I've addressed everything now.

src/provider_misc.jl
```diff
@@ -16,8 +16,8 @@ function process(r::JSONRPC.Request{Val{Symbol("initialize")},Dict{String,Any}},
         for (root, dirs, files) in walkdir(server.rootPath)
             for file in files
                 if splitext(file)[2]==".jl"
-                    filepath = joinpath(root, file)
-                    uri = string("file:///", replace(replace(filepath, '\\', '/'), ":", "%3A"))
+                    filepath = joinpath(root, file)
+                    uri = string("file:///", is_windows() ? replace(replace(filepath, '\\', '/'), ":", "%3A") : filepath)
```
ZacLN: still one too many slashes, on linux the filepaths lead with a forward slash so we end up with `file:////blah/blah`

davidanthoff: Ah, bugger. Ok, I've just pushed a new version, hopefully that fixes this for good.

PR #30, opened 2016-12-19
WIP: Caching

ZacLN:
- [x] Caches info from Base/Core in a simple `Dict` format and saves. 
- [x]  Builds cache automatically on a separate process.
- [x]  Handles all providers (hover/signature/definition/completion) through the cache.
- Increases startup time (loading cache).
- [x]  Follows simple `using` statements: adds module to cache, stores cache in a separate process, reloads cache on server.
- [x] Handle more complicated using statements (`using ModA:a, b`, `using ..ModA`, etc.)
- [x] Handle `import` statements.
- [x]  Follow `include` statements efficiently.
- [ ] Rejig `get_names` to amalgamate multiple function method definitions.
- [ ]Rejig `get_names` to amalgamate DataType/constructors.
- [x] Adds `tests_bruteforce.jl` which runs all providers across files in `src` but doesn't run by default.

This should be ready for review!


davidanthoff:
I briefly tried this and it does take a long time to build up the index... So I looked a bit more into extracting this kind of info. I *think* right now I can extract all functions, methods and their documentation from ``Base`` in about 0.6 seconds. I still need to add types and variables, but I'm getting there. Will require a bit more time to polish everything, and I hope I'm not missing something obvious. But if performance stays more like that, we could probably do without the on-disc cache, right? 

I would suggest that you leave this PR open for now, until we know whether my code actually stands a chance of working or not. Once I'm a bit further along and once we know whether my version is really that fast, we can decide what to do?

ZacLN:
Ok so I've had a look at this today and reformed the approach I was taking in #30. I've got it down to 2 seconds and ~10Mb for Base/Core by: 
- simplifying how methods are stored
- fixing the auto generated doc's (for variables without explicit documentation) to prevent the lookup of subtypes for concrete types.
- preventing the listing of subtypes for abstract variables.

With this timing there's no need to store it in a cache as it takes just as long to open and deserialize it.

davidanthoff:
Cool, that is great news! That essentially means I can close #32, and you continue on this, right?

> To be honest I feel like we're re-implementing a whole bunch of intrinsic Julia functionality here and may be better off simply spinning up a worker process onto which we import modules so as not to pollute the language server process.

I also just wanted to follow up on this comment from you in #32. Yes, I think we are re-implementing a lot of stuff, but on the other hand, I think if we really want a top notch solution we almost have to... I had thought about another process where we just load all the modules and then call the normal julia REPL completion function in that process whenever we get a completion request (and the same for help etc.), but I think the core problem is that we can't "unload" anything from that process, so if someone removes say a ``using`` statement in a file at the top, we would essentially have to kill that process, restart it again etc.... So my gut sense is that the approach here is still the right one. More work now, but I think the results should be much better, fingers crossed :)

In terms of process, I'll close my #32 now, and wait for you're new stuff, ok?

ZacLN:
Re-the child process approach- I don't think pollution is an issue as we can import modules and simply put access them through a layer of abstraction as we do now. The problem I ran into was that interprocess transfer seemed really slow

davidanthoff:
I don't think one can fully isolate imports. If a module adds a method to a function say from ``Base``, then that method is there to stay, right? But maybe that would be ok, not sure...

How did you do the inter-process communication?

ZacLN:
But a MethodTable can be filtered by `.module` right? This is what I (mis)meant by an abstraction layer.

I was just adding a worker process (i.e. `addprocs(1)`)

davidanthoff:
Yep, you are right, one could filter by ``.module``. So I guess then the one case that might still be problematic is if someone edits a file in a package (say in another editor) that is being used by a file in the current workspace. In theory we could add some filesystem modification watchers that catch that and then we might want to reload the information from that package. But I'm beginning to think that it might just be easier in that case to just restart the process where we load user modules and load everything from scratch... After all, that scenario is probably pretty rare.

Hm, strange about the communication speed, one would think that things should be fast with that. I'm not super familiar with the inter-process stuff, what functions did you use to send things back and for once the process was started?

ZacLN:
I've thought about it a bit more and implemented some different approaches.

1) Handling everything on the master leaves you affected by slow loading modules
2) Actual implementation of the faster caching only brings the (initial) caching down to 10sec though loading takes < 1sec.

3) Using a worker process (parallel) is a no go as you can only precompile (if needed) modules on node 1 which will be limiting.

davidanthoff:
So the package loading story suggests that we can't use the default julia ``addproc`` machinery to start the other process. But we can certainly just ``run`` it with the appropriate arguments, so that this child julia process actually uses the default package directory, right? At the end that is what we want: the LS process should use our own copy of the packages, but the process that extracts all the info from the packages should use the default julia package directory.

ZacLN:
Yep, this is what I've done [here](https://github.com/ZacLN/LanguageServer.jl/tree/cache2). (using the newer, faster caching). There's still a few kinks but they're internal data structure mismatches rather than major problems and I've been using it all week fairly successfully. 

At the moment I really see the caching approach as the best option

davidanthoff:
Excellent! I'm a bit under water right now so won't be able to look at this right now, I'll probably just wait until you have a version that could be merged.

Also, random other topic: we should consider submitting a talk to juliacon about this work here. Probably best discussed via email, do you want to send me an email and then we can figure that out? My email is on my github profile.

ZacLN:
Alright I'll tidy things up, merge onto this PR and let you know when it's ready to go.

ZacLN:
this is somewhat dependant on a [PR](https://github.com/JuliaLang/julia/pull/19964) to master being merged/back ported

ZacLN:
Ok I think this this is ready to be merged. the win64 failure looks unrelated

ZacLN:
@davidanthoff I think it may be worthwhile merging this now unless there's anything jumping out at you. Further improvements are most likely going to rely on the new parser which is probably 2-4 weeks away depending on other commitments.

src/cache.jl
```diff
@@ -0,0 +1,103 @@
...
+    end
+    return d
+end
+
+sig(x) = []
+function sig(x::Union{DataType,Function})
+    out = []
+    for m in methods(x)
+        p = string.(collect(m.sig.parameters[2:end]))
+        push!(out, (String(m.file), m.line, m.source.slotnames[2:length(p)+1], p))
```
davidanthoff: This line crashes for me. I can get rid of the first crash by using ``string(m.file)`` instead of ``String(m.file)``, but then I get another crash
````
ERROR: type Method has no field source
 in sig(::Function) at C:\Users\anthoff\.julia\v0.5\LanguageServer\src\cache.jl:50
 in modnames(::Module, ::Dict{Any,Any}) at C:\Users\anthoff\.julia\v0.5\LanguageServer\src\cache.jl:38
 in eval(::Module, ::Any) at .\boot.jl:234
 in process_options(::Base.JLOptions) at .\client.jl:239
 in _start() at .\client.jl:318
````
I guess this is some julia 0.5 vs 0.6 error?

File: src/cache.jl [hunk 2]
```diff
@@ -0,0 +1,103 @@
...
+
+
+
+
+
+updatecache(absentmodule::Symbol, server) = updatecache([absentmodule], server)
+
+function updatecache(absentmodules::Vector{Symbol}, server)
+    send(Message(3, "Adding $(join(absentmodules, ", ")) to cache, this may take a minute"), server)    
+    run(`julia -e "using LanguageServer;delete!(Base.ENV, \"JULIA_PKGDIR\"); top = LanguageServer.loadcache(); for m in [$(join((m->"\"$m\"").(absentmodules),", "))]; LanguageServer.modnames(m, top); end; LanguageServer.savecache(top)"`)
```
davidanthoff: This should be ``run(`$JULIA_HOME/julia`` etc. to make sure we are calling the same julia binary here.

src/provider_misc.jl
```diff
@@ -26,6 +26,13 @@ function process(r::JSONRPC.Request{Val{Symbol("initialize")},Dict{String,Any}},
     end
     response = JSONRPC.Response(get(r.id), InitializeResult(serverCapabilities))
     send(response, server)
+
+    if isempty(server.cache)
+        send(Message(3, "Building cache, this may take a few minutes"), server)
+        run(`julia -e "using LanguageServer; top = Dict();LanguageServer.modnames(Main, top); LanguageServer.savecache(top)"`)
```
davidanthoff: Same here, this should be ``run(`$JULIA_HOME/julia`` etc. to make sure we are calling the same julia binary here.

ZacLN:
does this work now?

davidanthoff:
I'll try to find some time tomorrow to check it out. Completely under water with other work right now...

ZacLN:
Cheers

davidanthoff:
It seems to work now! Great!

I have a whole bunch of follow up things, but probably easier to do those in new PRs on master, so I'll merge this once the tests pass.

PR #32, opened 2016-12-24
WIP Fast base and package information extraction

davidanthoff:
@ZacLN This is a complete mess right now, just thought I'd let you see where I am :)

### DIFF
--- a/src/base_package_info.jl
+++ b/src/base_package_info.jl
@@ -1,0 +1,129 @@
+
+module ModuleA
+    importall Distributions
+
+    module ModuleB
+        bar()=3
+    end
+    
+    """
+    foo1 docs
+    """
+    foo1()=bar()
+
+    """
+    foo2 function docs
+    """
+    function foo2 end
+
+    """
+    foo2 method docs
+    """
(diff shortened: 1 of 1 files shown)


ZacLN:
Sorry @davidanthoff I've been away in a wifi-less land for a few weeks so haven't had a chance to look at this.

This is definitely speedy but can't quite tell if that's because it's not pulling the same amount of information. I don't know how far you delved into my implementation but a few things that I thought were key and a few new things:

- storing modules/submodules in strictly separate `Dict`s with a list of exported names. This  makes completions much easier/faster.
- Maybe make use of some stuff in `DataStructures` e.g. `SortedDicts`? 
- At the moment you're not getting all the relevant docs (from a user's perspective). I can't seem to get an entry for `sin` caching `Base` or `Base.Math` ( I may well be missing something here).
- It _may_ be useful to `MultiDoc`s separately so that we can supply the specifically relevant entry when possible.

To be honest I feel like we're re-implementing a whole bunch of intrinsic Julia functionality here and may be better off simply spinning up a worker process onto which we import modules so as not to pollute the language server process

PR #38, opened 2017-02-24
WIP: Load cache in spawned process

ZacLN:
This loads a child process for build the cache for `Main` and any references mods. There's some sort of conflict between the user's package directory and the LS one but  I can't quite pin down the problem yet.

I still want to get the `Base` caching faster so I'm going to look into transferring it submodule by submodule rather than as a single object

### DIFF
--- a/src/cache.jl
+++ b/src/cache.jl
@@ -95,7 +95,7 @@ updatecache(absentmodule::Symbol, server) = updatecache([absentmodule], server)
 
 function updatecache(absentmodules::Vector{Symbol}, server)
     fname = functionloc(updatecache, Tuple{Symbol, Any})[1]
-    o,i, p = readandwrite(`julia -e "delete!(Base.ENV, \"JULIA_PKGDIR\");
+    o,i, p = readandwrite(`$JULIA_HOME/julia -e "delete!(Base.ENV, \"JULIA_PKGDIR\");
     include(\"$fname\");
     top=Dict();
     for m in [$(join((m->"\"$m\"").(absentmodules),", "))];
--- a/src/cache.jl
+++ b/src/cache.jl
@@ -94,10 +94,24 @@ end
 updatecache(absentmodule::Symbol, server) = updatecache([absentmodule], server)
 
 function updatecache(absentmodules::Vector{Symbol}, server)
-    send(Message(3, "Adding $(join(absentmodules, ", ")) to cache, this may take a minute"), server)    
-    run(`$JULIA_HOME/julia -e "using LanguageServer;delete!(Base.ENV, \"JULIA_PKGDIR\"); top = LanguageServer.loadcache(); for m in [$(join((m->"\"$m\"").(absentmodules),", "))]; LanguageServer.modnames(m, top); end; LanguageServer.savecache(top)"`)
-    server.cache = loadcache()
-    send(Message(3, "Cache stored at $(joinpath(Pkg.dir("LanguageServer"), "cache", "docs.cache"))"), server)
+    fname = functionloc(updatecache, Tuple{Symbol, Any})[1]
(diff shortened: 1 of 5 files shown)


src/cache.jl
```diff
@@ -94,10 +94,24 @@ end
 updatecache(absentmodule::Symbol, server) = updatecache([absentmodule], server)
 
 function updatecache(absentmodules::Vector{Symbol}, server)
-    send(Message(3, "Adding $(join(absentmodules, ", ")) to cache, this may take a minute"), server)    
-    run(`$JULIA_HOME/julia -e "using LanguageServer;delete!(Base.ENV, \"JULIA_PKGDIR\"); top = LanguageServer.loadcache(); for m in [$(join((m->"\"$m\"").(absentmodules),", "))]; LanguageServer.modnames(m, top); end; LanguageServer.savecache(top)"`)
-    server.cache = loadcache()
-    send(Message(3, "Cache stored at $(joinpath(Pkg.dir("LanguageServer"), "cache", "docs.cache"))"), server)
+    fname = functionloc(updatecache, Tuple{Symbol, Any})[1]
+    o,i, p = readandwrite(`julia -e "delete!(Base.ENV, \"JULIA_PKGDIR\");
```
davidanthoff: This needs ``$JULIA_HOME/julia``.

src/provider_misc.jl
```diff
@@ -27,11 +27,18 @@ function process(r::JSONRPC.Request{Val{Symbol("initialize")},Dict{String,Any}},
     response = JSONRPC.Response(get(r.id), InitializeResult(serverCapabilities))
     send(response, server)
 
-    if isempty(server.cache)
-        send(Message(3, "Building cache, this may take a few minutes"), server)
-        run(`$JULIA_HOME/julia -e "using LanguageServer; top = Dict();LanguageServer.modnames(Main, top); LanguageServer.savecache(top)"`)
-        server.cache = loadcache()
-        send(Message(3, "Cache stored at $(joinpath(Pkg.dir("LanguageServer"), "cache", "docs.cache"))"), server)
+    o,i, p = readandwrite(`julia -e "using LanguageServer;
```
davidanthoff: This needs ``$JULIA_HOME/julia``.

davidanthoff:
> This loads a child process for build the cache for Main and any references mods. There's some sort of conflict between the user's package directory and the LS one but I can't quite pin down the problem yet.

Maybe an idea might be this: we add another field to the LS instance type ``user_pkg_path``. in the ``main.jl`` file in the VS Code extension we pass the value of ``Base.ARGS[1]`` for that one (which is always the path of the user pkg dir). Then, we *always* set JULIA_PKGDIR to that path for the client process that we start for creating the cache. In that way the client process will a) always access the user package folder and b) always use the user precompile folder, which also should make things faster. Of course that client process then doesn't know about the ``LanguageServer`` package, but I think you already modified the code such that the client process actually never has ``using LanguageServer`` in it, and instead just ``include``s whatever code it needs from the LanguageServer. I think the main benefit here would be that we would reuse the user package folder precompile cache. I suspect that ``delete!(Base.ENV, \"JULIA_PKGDIR\")`` you have right now doesn't change the precompile location and might mess things up.

davidanthoff:
Oh, and to modify ``JULIA_PKGDIR`` for the child process, you'll have to use an explicit ``Cmd`` instance to spawn the child process, there is an option with those to set the environmental variables for the new child process.

davidanthoff:
> I still want to get the ``Base`` caching faster so I'm going to look into transferring it submodule by submodule rather than as a single object

Oh, another idea on that one: if I recall my perf testing, then the thing that really takes time is extracting the method signatures, whereas getting the doc strings even for all of Base is super fast. So one could maybe even split this up, first transfer the doc strings back, and then in a second step the method signatures. That might cut down on the time until at least doc strings are available.

PR #40, opened 2017-03-23
Use base64 encoding for communication

davidanthoff:
I couldn't get things to work without this on my machine...

### DIFF
--- a/src/cache.jl
+++ b/src/cache.jl
@@ -94,17 +94,22 @@ end
 updatecache(absentmodule::Symbol, server) = updatecache([absentmodule], server)
 
 function updatecache(absentmodules::Vector{Symbol}, server)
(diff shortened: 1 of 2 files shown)


ZacLN:
Hmm, this is causing me problems now:
```
ERROR (unhandled task failure): ArgumentError: incorrect base64 format, block must be at least 2 and at most 4 bytes
```
when doing `using Parser`. 

Seems to be fine with other packages, any idea what could cause this?

davidanthoff:
Hm, very strange... I would think that base64 encoding doesn't use unicode?

ZacLN:
I have no idea how any of this works tbh, I was just thinking that some of the unicode symbols take up strange sized data block

PR #41, opened 2017-03-30
add child process lock

ZacLN:
Previously opening a file with a few imports and making language server requests would spin up child processes each time a request was made

### DIFF
--- a/src/cache.jl
+++ b/src/cache.jl
@@ -99,7 +99,6 @@ function updatecache(absentmodules::Vector{Symbol}, server)
 
 
     o,i, p = readandwrite(Cmd(`$JULIA_HOME/julia -e "include(\"packages/LanguageServer/src/cache.jl\");
-    delete!(Base.ENV, \"JULIA_PKGDIR\");
     top=Dict();
     for m in [$(join((m->"\"$m\"").(absentmodules),", "))];
         modnames(m, top); 
--- a/src/cache.jl
+++ b/src/cache.jl
@@ -94,10 +94,11 @@ end
 updatecache(absentmodule::Symbol, server) = updatecache([absentmodule], server)
 
 function updatecache(absentmodules::Vector{Symbol}, server)
-    server.cp_lock && return
+    env_new = copy(ENV)
+    env_new["JULIA_PKGDIR"] = server.user_pkgdir
 
-    server.cp_lock = true
-    o,i, p = readandwrite(`$JULIA_HOME/julia -e "include(\"packages/LanguageServer/src/cache.jl\");
+
+    o,i, p = readandwrite(Cmd(`$JULIA_HOME/julia -e "include(\"packages/LanguageServer/src/cache.jl\");
     delete!(Base.ENV, \"JULIA_PKGDIR\");
     top=Dict();
     for m in [$(join((m->"\"$m\"").(absentmodules),", "))];
@@ -108,7 +109,7 @@ function updatecache(absentmodules::Vector{Symbol}, server)
     serialize(io_base64, top);
     close(io_base64);
     str = takebuf_string(io);
-    println(STDOUT, str)"`)
+    println(STDOUT, str)"`, env=env_new))
     
     @async begin 
         str = readline(o)
@@ -120,6 +121,5 @@ function updatecache(absentmodules::Vector{Symbol}, server)
                 server.cache[k] = mods[k]
             end
         end
-        server.cp_lock = false
     end
 end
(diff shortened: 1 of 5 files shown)


src/cache.jl
```diff
@@ -94,7 +94,11 @@ end
 updatecache(absentmodule::Symbol, server) = updatecache([absentmodule], server)
 
 function updatecache(absentmodules::Vector{Symbol}, server)
+    server.cp_lock && return
+
+    server.cp_lock = true
     o,i, p = readandwrite(`$JULIA_HOME/julia -e "include(\"packages/LanguageServer/src/cache.jl\");
+    delete!(Base.ENV, \"JULIA_PKGDIR\");
```
ZacLN: @davidanthoff, I seem to need this to be able to search my user package dir, is this specific to me/linux?

src/cache.jl
```diff
@@ -94,7 +94,11 @@ end
 updatecache(absentmodule::Symbol, server) = updatecache([absentmodule], server)
 
 function updatecache(absentmodules::Vector{Symbol}, server)
+    server.cp_lock && return
```
davidanthoff: Is this how we want things to be? Essentially if you can't acquire the lock, give up? That seems not ideal... Generally, a better strategy might be to have just one background green task that waits on a ``Channel`` for requests to update the cache with a given module, and then that task starts the child process. So the normal language server code would no longer start a new child process when it needs stuff for a module, instead it would push that module name into the ``Channel``, then the background task processes the items from the channel one by one, i.e. we would never have more than one julia child process running at the same time that loads modules, but at the same time all requests would be served eventually.

src/cache.jl
davidanthoff: Hm, there might be a Windows/Unix difference here... The current theory is this: VS Code starts the language server and sets its ``JULIA_PKGDIR`` to a folder under the VS Code extension. That results in all the precompile files being there, separate from the normal user location.

At least on windows, the child process that we start here in code does not inherit that environmental variable setting (I'm 99% sure), instead the child process has ``JULIA_PKGDIR`` point towards the normal user package location, which is what we want.

I'll prep a new PR that makes sure our child processes always have the right value for ``JULIA_PKGDIR`` there, should be fairly simple. I don't think deleting ``JULIA_PKGDIR`` is a good idea, because then things like the location for precompile etc. are really left unclear. Instead, we should start the child process with the correct value for that env variable.

ZacLN: I think your bang on the money, linux inherits the imposed ENV variables.

src/cache.jl
ZacLN: In practice 'give up' means: next time that a key is pressed, the mouse hovers etc. then we'll see if we can launch a new child process. 

I've found package loading to be an action ripe for segfault-y crashing (build errors etc.) and a julia process can really take up a bit too much RAM which is why I don't particularly fancy a persistent child process.

In our super new Parser I've got a workspace/directory parsing system that tracks all module use in a more centralised manner which also allows us to track packages that refuse to load.

davidanthoff: Ah, I didn't mean a persistent child process, I just meant a persistent thread in the main ls process that essentially works on a queue of tasks to process, and then still starts individual julia child processes. It would just serialize the whole thing.

ZacLN: Ok, I've not meddled with `Task`s hence my confusion, I'll have a fiddle in the morning by adding a `Channel` to the `LanguageServerInstance`

ZacLN: does this pseudo code do what you were meaning?
```
c = Channel(32)
while true # main loop in run at languageserverinstantce.jl line 29
    # call to get_names(Val{import/using/importall}) at namespace.jl lines 168, 187
    put!(server.user_modules)

    if isready(c)
        missing_mod = take!(server.usermodules)
        if !(missing_mod in keys(server.cache))
            updatecache(missing_mod)
        end
    end
end
```

davidanthoff: What I had in mind was something like this:
````julia
ch = Channel()
@schedule begin
    for i in ch
        if !(missing_mod in keys(server.cache))
            updatecache(missing_mod)
        end
    end
end
````
This code would start when the language server starts, and then just keeps running as a background thread. I think it will "wake up" whenever there is something in ``ch`` to be processed. Also, when ``updatecache`` calls ``readandwrite``, i.e. starts the child process, this background thread should "fall asleep" again, and other tasks can continue processing stuff.

And then in the normal code, if one wants to add a given module to the cache, one just does
````julia
put!(ch, modulename)
````
or something like that.

ZacLN:
Alright this is set up now, I'll fix the conflicts after #42 is fixed to pass tests (we either need to add `ENV["JULIA_PKGDIR"] = Pkg.dir()` to the start of `runtests.jl` or could set it more generally which may help with debugging)

ZacLN:
This should be good to go, I've add a per-session blacklist for modules that fail to load

PR #45, opened 2017-04-04
Move linter into separate process

davidanthoff:
This is not perfect, but given that we want to kick out Lint.jl hopefully soon I think it might be good enough for now?

This no longer uses a Lint package that we bring with the VS extension, instead is requires Lint to be installed in the user pkg directory. I think that is the only way right now to have the lint process use the user precompiled binaries and also really isolate our LS process from any bugs/errors/problems in the lint process.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
 julia 0.5
 JSON 0.8.0
-Lint 0.2.5
+Lint 0.5.0
 URIParser 0.1.6
--- a/src/LanguageServer.jl
+++ b/src/LanguageServer.jl
@@ -1,7 +1,6 @@
 module LanguageServer
 
 using JSON
-using Lint
 using URIParser
 using JuliaParser
 
--- a/src/languageserverinstance.jl
+++ b/src/languageserverinstance.jl
@@ -9,6 +9,7 @@ type LanguageServerInstance
 
     debug_mode::Bool
     runlinter::Bool
+    linter_is_installed::Bool
 
     user_pkg_dir::String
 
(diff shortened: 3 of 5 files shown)


davidanthoff:
@ZacLN I think the test failure happens because I moved the whole send message story into another task. I need to do that because otherwise multiple tasks might write to STDOUT, and that doesn't work because the messages can get mangled. Not sure how to fix the hover test now, though... Any idea?

ZacLN:
I haven't had a chance to have a look at this yet but I'll try and get my head round it tomorrow morning and give my 2p!

ZacLN:
Hmm, I couldn't get any where with this. Is it just the tests failing for you? I can't get it running at all, it can't open the pipe to communicate

davidanthoff:
Yeah, for me it is just the tests that fail. I have a Linux system, so I'll try to get it to run there first and then will ping you again.

davidanthoff:
@ZacLN Should we just nix this? There are so many false positives with this in any case, so lets just move over to your diagnostics with the next release and completely drop Lint.jl?

ZacLN:
I think so. I've added everything from the [list](https://lintjl.readthedocs.io/en/latest/features/) of features that can be implemented on the run through of the parser (i.e. only local scope knowledge). 

A second pass function should be able to collect most conflicts deriving from undeclared/misused variables/functions/types

davidanthoff:
Ok, I'll leave the PR open for now, but fingers crossed we will just close it without merging once you have merged your new stuff :)

PR #49, opened 2017-04-07
Fix base64 encoding bug

davidanthoff:
This should fix the base64 encoding bugs we were still seeing.

### DIFF
--- a/src/cache.jl
+++ b/src/cache.jl
@@ -113,7 +113,7 @@ function updatecache(absentmodules::Vector{Symbol}, server)
     
     @async begin 
         str = readline(o)
-        data = base64decode(str)
+        data = base64decode(chomp(str))
         mods = deserialize(IOBuffer(data))
         for k in keys(mods)
             if !(k in keys(server.cache))
--- a/src/provider_misc.jl
+++ b/src/provider_misc.jl
@@ -45,7 +45,7 @@ function process(r::JSONRPC.Request{Val{Symbol("initialize")},Dict{String,Any}},
 
     @async begin
         str = readline(o)
-        data = base64decode(str)
+        data = base64decode(chomp(str))
         mods = deserialize(IOBuffer(data))
         for k in keys(mods)
             if !(k in keys(server.cache))


ZacLN:
this is giving me 
```
ERROR: MethodError: no method matching SubArray{UInt8,1,Array{UInt8,1},Tuple{UnitRange{Int64}},true}()
Closest candidates are:
  SubArray{UInt8,1,Array{UInt8,1},Tuple{UnitRange{Int64}},true}{T,N,P,I,L}(::Any, ::Any, ::Any, ::Any) at subarray.jl:14
  SubArray{UInt8,1,Array{UInt8,1},Tuple{UnitRange{Int64}},true}{T}(::Any) at sysimg.jl:24
Stacktrace:
 [1] close(::Base.AbstractIOBuffer{SubArray{UInt8,1,Array{UInt8,1},Tuple{UnitRange{Int64}},true}}) at ./iobuffer.jl:251
 [2] base64decode(::SubString{String}) at ./base64.jl:234
```

is this a v0.5/6 issue?

src/cache.jl
```diff
@@ -113,7 +113,7 @@ function updatecache(absentmodules::Vector{Symbol}, server)
     
     @async begin 
         str = readline(o)
-        data = base64decode(str)
+        data = base64decode(chomp(str))
```
ZacLN: `base64decode` doesn't work with a SubString, 
`data = base64decode(String(chomp(str)))`
gets this working. We reallocated the full string

davidanthoff: I think that is actually a bug on julia ``master``, see JuliaLang/julia#21314. Maybe we'll wait and see what they say over there? If they fix that quickly, we can wait here with the merge, otherwise we can go with the copy approach?

ZacLN: Yep that's fine w/ me

davidanthoff: Alright, I think Jeff fixed this on ``master``, can you maybe confirm and then merge this?

davidanthoff:
I can close this, right? We are not using any of this in the current version?

ZacLN:
Yep

PR #51, opened 2017-04-14
Use JSON.lower specialization

TotalVerb:
`JSON._writejson` was an internal implementation detail and should not have been used to override the serialization of `Range`. In JSON 0.9.0, `_writejson` has been removed.

cc @tkelman, see https://github.com/JuliaLang/METADATA.jl/pull/8820#issuecomment-294069652

### DIFF
--- a/src/protocol.jl
+++ b/src/protocol.jl
@@ -19,12 +19,8 @@ type Range
     stop::Position # mismatch between vscode-ls protocol naming (should be 'end')
 end
 
-function JSON._writejson(io::IO, state::JSON.State, a::Range)
-    Base.print(io,"{\"start\":")
-    JSON._writejson(io,state,a.start)
-    Base.print(io,",\"end\":")
-    JSON._writejson(io,state,a.stop)
-    Base.print(io,"}")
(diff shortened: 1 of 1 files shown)


davidanthoff:
@tkelman If we merge this into ``master`` it won't affect the tagged version and won't help with package evaluator runs (which I assume triggered this?) because we won't tag anything off ``master`` for quite a while. Do you want me to create a bug fix branch off the last tagged version and then tag a patch release with this fix there? I assume that would help with package evaluator runs?

tkelman:
We've done https://github.com/JuliaLang/METADATA.jl/issues/8820 in metadata for now which means if you have LanguageServer 0.0.1 installed, JSON will be held back to 0.8.x. In order to have LanguageServer be able to coexist with JSON 0.9.0 you may want to make a backport release, sure.

davidanthoff:
Alright, lets not do a backport release. I can't imagine that anyone is using this package in any case, except us. I do want to make sure package evaluator isn't blocked by this issue, but with the bounds in METADATA.jl that should be achieved already.

Issue #11, opened 2016-11-28
How to handle packages that are loaded from user code?


davidanthoff:
Ideally we would not ever load any user code into the LS process. This makes it non-trivial to get at documentation, symbols, methods etc. from packages that user code might load. I'm not entirely sure how to best handle this, here are some rough ideas:

1. span a new julia process, that julia process loads the package, and returns all the necessary information to the LS process, say in JSON format. This could be done async, so maybe some of the symbols and info from the package would only show up after a while.
2. run our static analysis on the package that a user loads. This will probably miss some things...

In general it might also make sense to cache symbol/method/function/doc information about packages on file or something...

ZacLN:
Option 1 is the way to go I think, static analysis is going to miss too much loop/eval'd code generation stuff. 

Caching it all make sense and we can check for changes with whatever `
compilecache` uses

davidanthoff:
I guess one question is: how can we get the list of all *methods* defined in a given module. In particular, can we somehow get methods that actually extend a function in say base, that were defined in a module?

ZacLN:
We can use the static analysis stuff and find all import statements, then filter the relevant functions' method tables by their  `.module`

davidanthoff:
Ah, cool! And actually, we might just see how performance looks like if we just look at all methods and check their ``.module``, if that is not too slow we could avoid the whole static analysis, right?

ZacLN:
I've whipped up a simple [script](https://gist.github.com/ZacLN/111ffa8ee7cc28ac863de02835dba086) that pulls hovers and goto defs from packages. It only takes ~ 5 seconds on my computer to pull everything from `Base`, this will be a bit longer when I add signatures but I don't think it's anything to worry about. Size for `Base` is around 3Mb so storage is no great ordeal

davidanthoff:
Do you mean line 4 takes 5 secs on your system? That line takes 0.1 seconds on my system :) But yes, speed doesn't seem to be a huge issue here.

ZacLN:
Well that's distressing!

ZacLN:
Oh no I meant to pull all the documents from running `cachemodule`! I was worried my computer was busted somehow...

Anyway I've fiddled with this and have it working to pull `Base`/`Core` and all sub modules and to cache user installed packages.

[...]

PR #58, opened 2017-05-25
Updates LS for CSTParser refactor

ZacLN:
Changes: 
- Uses new CSTParser format for expressions.
- Improved accuracy of scoping.
- Handles imports of individual symbols.
- Adds lint-package command to check REQUIRE against imported modules.
- Adds start of framework for on-type linting.


src/languageserverinstance.jl
```diff
@@ -26,13 +23,15 @@ function send(message, server)
...
     @schedule begin
-        for missing_module in server.user_modules
-            if !(missing_module in keys(server.cache)) && !(missing_module in wontload_modules)
-                updatecache(missing_module, server)
-                if !(missing_module in keys(server.cache))
-                    push!(wontload_modules, missing_module)
+        for (modname,uri,loc) in server.user_modules
+            if !(modname in wontload || modname in loaded) 
+                try 
+                    @eval import $modname
```
davidanthoff: Alright, lets do this for now, but I would still like to get rid of this eventually...

ZacLN: I agree with your thoughts in #11, it's just that it's much easier to add the features and find out whats possible having everything 'just there' rather than as a three step process of retrieve/store/transfer.

davidanthoff: Yeah, lets move forward and try to solve this later!

PR #73, opened 2017-05-31
move all parsing inside try block, return parse error diagnostic on serious parsing failure

ZacLN:


davidanthoff:
Hm, I'm not a fan of this... Essentially this will handle bugs in the parser code that we should just fix, right? I've always felt that it is better to just let things crash in this case, get a bug report from users and fix things. With this, I'm pretty sure we will just never learn about those kinds of bugs...

ZacLN:
Alright I'll give it a go removing the try block completely

davidanthoff:
Yes, that would be great! Sorry for being a pain on this point, but I hope it will pay off longterm.

ZacLN:
No you're right, it should be fine now anyway

PR #75, opened 2017-05-31
Lint pass

ZacLN:
Adds a lint pass after parsing, checks for use of undeclared variables.

### DIFF
--- a/src/document.jl
+++ b/src/document.jl
@@ -7,9 +7,10 @@ type Document
     code::CSTParser.File
     diagnostics::Vector{CSTParser.Diagnostics.Diagnostic}
     _version::Int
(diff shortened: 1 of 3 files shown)


ZacLN:
This is working well, is `julia-vscode` tracking the latest version of CSTParser or master? If the former I'll push a new version so that this can be tested

davidanthoff:
It is tracking ``master`` of CSTParser, but one needs to manually do the tracking. I just pushed a new version of julia-vscode that uses the latest CSTParser. In general, julia-vscode is completely isolated from METADATA for all packages.

This PR here is ready for review, right?

ZacLN:
Yep this guy's ready to go

PR #76, opened 2017-06-01
fix lint-package path

ZacLN:
I'm not sure if a similar change needs to be made on the windows side?

### DIFF
--- a/src/lint.jl
+++ b/src/lint.jl
@@ -2,7 +2,7 @@ function process(r::JSONRPC.Request{Val{Symbol("julia/lint-package")},Void}, ser
     warnings = []
     if isdir(server.rootPath) && "REQUIRE" in readdir(server.rootPath)
         topfiles = []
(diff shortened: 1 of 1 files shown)


src/lint.jl
```diff
@@ -2,7 +2,7 @@ function process(r::JSONRPC.Request{Val{Symbol("julia/lint-package")},Void}, ser
     warnings = []
     if isdir(server.rootPath) && "REQUIRE" in readdir(server.rootPath)
         topfiles = []
-        rootUri = is_windows() ? string("file:///", replace(replace(server.rootPath, "\\", "/"), ":", "%3A")) : string("file://", server.rootPath, "/src")
+        rootUri = is_windows() ? string("file:///", replace(replace(server.rootPath, "\\", "/"), ":", "%3A")) : joinpath("file://", server.rootPath, "src")
```
davidanthoff: Probably like this:
````julia
string("file:///", replace(joinpath(replace(server.rootPath, "\\", "/"), "src"), ":", "%3A"))
````

PR #81, opened 2017-06-08
fix diagnostics positioning, add tests

ZacLN:


ZacLN:
@davidanthoff , I'm at a bit of a loss here. This passes on my setup and equivalent tests directly on [CSTParser](https://github.com/ZacLN/CSTParser.jl/pull/16) give the correct output. Could you test on your system whether the following tests fail:
```
using CSTParser
using Base.Test

ps = ParseState("""
abstract T
type T end
immutable T end
typealias T T
bitstype T 8""")
x,ps = CSTParser.parse(ps, true)
@test ps.diagnostics[1].loc == 0:8
@test ps.diagnostics[2].loc == 11:15
@test ps.diagnostics[3].loc == 22:31
@test ps.diagnostics[4].loc == 38:47
@test ps.diagnostics[5].loc == 52:60
```

davidanthoff:
Isn't the CI build just using an old CSTParser version? The last tagged version of CSTParser is v0.0.5, so when travis runs it is missing all of [these](https://github.com/ZacLN/CSTParser.jl/compare/v0.0.5...master) commits.

ZacLN:
Ah yes of course, Doh! I'll sort this then we're ready

PR #84, opened 2017-06-13
Add `end` completion

ZacLN:


andyferris:
Will this decrement indentation of `end` automatically as I type it?

ZacLN:
At the moment in vscode de-indenting occurs on hitting enter (after `end`, `else`, etc.)

andyferris:
OK, I see that now. This will become more obvious after this fix :)

davidanthoff:
@andyferris If you want to add another voice over at the VS Code repo to fix this properly, here is the issue: Microsoft/vscode#2272.

andyferris:
Done.

PR #85, opened 2017-06-21
fix return tuple of parse_jmd

ZacLN:


davidanthoff:
I assume this should be included for our 0.6 release, right?

PR #86, opened 2017-06-22
transfer full doc text on change

ZacLN:
This should fix [this](https://github.com/JuliaEditorSupport/julia-vscode/issues/188), and incremental text update is not needed until incremental parsing returns

### DIFF
--- a/src/provider_misc.jl
+++ b/src/provider_misc.jl
@@ -1,5 +1,5 @@
 const serverCapabilities = ServerCapabilities(
-                        TextDocumentSyncKind["Incremental"],
+                        TextDocumentSyncKind["Full"],
                         true, #hoverProvider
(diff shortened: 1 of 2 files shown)


davidanthoff:
Should we publish a bug fix release 0.6.1 with this?

ZacLN:
Yep, that bug was pretty major

davidanthoff:
There is still something wrong. When I open a file now and type, it moves all the green underlines for diagnostics in the rest of the file around while I type.

ZacLN:
I can't work out what's getting confused here and I'm not going to have enough time to dig into it for a few days, Want to revert this so that at least master works as before?

davidanthoff:
Alright, I revert this for now.

PR #89, opened 2017-06-25
revert PR86

ZacLN:
Reverts use of full document syncing just so that `master` works


ZacLN:
Didn't see the reversion had already happened, closing

PR #102, opened 2017-07-03
update readme

ZacLN:
Add reference to [atom-julia-lsp-client](https://github.com/pfitzseb/atom-julia-lsp-client)


ZacLN:
@pfitzseb set it up in about an hour!

PR #103, opened 2017-07-03
Added instructions for setting up neovim

joshbode:


davidanthoff:
Excellent, this is really helpful! One thought: If we add all the detailed instructions/manuals for various editors to the README it will get quite long... What about this: we add a Wiki page for each editor here in this repo that has all the detailed instructions for that editor (i.e. most of the text that is in this PR), and then in the README we just add one short section that has links to all the Wiki pages for the various editors?

joshbode:
I think that makes sense - is the wiki activated?
Seems to just jump back to the main repo page.

davidanthoff:
It should work, I just added a page for Neovim, and I think you should be able to edit that?

joshbode:
Thanks - I can see it now, but can't see an edit button for the page.

davidanthoff:
Ah, wrong security setting. I think you should be able to edit it now, I just changed it.

joshbode:
Ta! Saved, and the README has been adjusted, too.

davidanthoff:
Excellent, thanks!

joshbode:
No problems! I'm excited to try this out in nvim :)

PR #140, opened 2017-07-29
tests pass against JSON 0.8.1

tkelman:


davidanthoff:
Test coverage is not really high enough to rely on it to justify this change. The only way to test this change would be to build the VS Code extension with JSON v0.8.1 and then manually test whether it still works, but that is a fair bit of work.

tkelman:
what new feature are you using in JSON that justifies requiring 0.9?

tkelman:
As far as I can tell, you aren't relying on any of the serialization-related changes from https://github.com/JuliaIO/JSON.jl/compare/v0.8.1...v0.9.0. The VS Code extension is built independently right, so this only matters for people using LanguageServer via Julia code and Pkg. Pinning JSON to 0.8.1-0.8.3 shouldn't force a downgrade of LanguageServer to 0.0.1 or Pkg resolver errors.

davidanthoff:
@ZacLN I believe you bumped the JSON dep to v0.9.0, do you remember why?

@tkelman Yes, none of this matters for the VS Code extension, we don't use the package manager for that deployment.

tkelman:
prior to https://github.com/JuliaEditorSupport/LanguageServer.jl/pull/51 this package had been using `_writejson` which was internal and removed in JSON 0.9.0, but its proper replacement `JSON.lower` was added earlier, in 0.7.0 https://github.com/JuliaIO/JSON.jl/commit/495c75ff122abcac409cc79b7c4bc7d005872d7a

tkelman:
ping @ZacLN

ZacLN:
It's taking an eternity to drop my JSON to 0.8.1 (like 10 minutes) so I'm going to check that it runs tomorrow but as far as I can tell it should be fine.

I think I got confused/wasn't paying attenion when #51 was made thinking the JSON api had changed, i

tkelman:
How is it taking 10 minutes to run Pkg.pin ?

ZacLN:
Dunno, maybe github was being funky?
Pkg.checkout/status often takes around 5 mins for me

davidanthoff:
You need a new computer with a SSD :)

ZacLN:
I've been hunting for a non-stupidly expensive/large quad-core laptop for about a year... I can but dream

tkelman:
Pkg.pin shouldn't need network access at all?

ZacLN:
well I have no idea what was going on then, it left me w/ a mangled JSON install

tkelman:
You can try things in a clean separate JULIA_PKGDIR if that makes it easier

tkelman:
Verdict? That's 4 proposed tags with a too-tight requirement now

davidanthoff:
I guess the reality is that neither @ZacLN nor I are willing to spend time to properly test less tight version bounds. So I guess the question for the METADATA maintainers is whether you think it is overall more beneficial to the user base to have these tags registered with tight version bounds or not have them in METADATA. From my point of view I don't really care that much, given that we don't use the package manager for the VS Code extension deployment story.

tkelman:
What does "properly test" entail (and how would contributors meet that on more involved changes)? There are use cases where leveraging this code through Julia and Pkg would be worthwhile. But if it doesn't require JSON 0.9 to function correctly, which from looking at Pkg.test, your code and JSON.jl's code it almost certainly doesn't, then shouldn't force that and disallow JSON 0.8.1+ if someone has a good reason to be using that.

ZacLN:
I pinned at 0 8.1 today and everything is working, can we just get to our final v0.7 release before dropping this ¿

tkelman:
I don't know what and where you want to be releasing things, if attobot should be disabled on some of the tags here or what? Since this applies to all the pending-on-metadata tags from what I can tell, I can easily go make this same change on the attobot branches for each of them if you agree with that.

ZacLN:
I've tested this with v0.8.1, it works. I'm happy for the downgrade to go forward

PR #148, opened 2017-08-01
reparse files on change of git branch

ZacLN:
Who knew about `watch_file` ? 
This reparses/lints the workspace when you change branch. This seems to work but I don't know enough about git's internals to know if this is the best way to implement it

### DIFF
--- a/src/provider_misc.jl
+++ b/src/provider_misc.jl
@@ -37,6 +37,13 @@ end
 
 
 function process(r::JSONRPC.Request{Val{Symbol("initialized")},Dict{String,Any}}, server) 
+    if isdir(joinpath(server.rootPath, ".git")) && isfile(joinpath(server.rootPath, ".git", "HEAD"))
+        @async begin
+            watch_file(joinpath(server.rootPath, ".git", "HEAD"))
+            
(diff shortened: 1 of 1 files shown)


davidanthoff:
Hm, do we really need this? I think [this](https://github.com/JuliaEditorSupport/julia-vscode/blob/master/src/extension.ts#L252) should trigger a [DidChangeWatchedFiles](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#workspace_didChangeWatchedFiles) notification for any ``.jl`` file, so maybe we need to add more logic to the [handler](https://github.com/JuliaEditorSupport/LanguageServer.jl/blob/74223a0cb8d6a55e4c019cdc6673b4145519add3/src/provider_misc.jl#L134) for that? In general, the model seems to be that VS Code monitors these files and then sends notification events, and we probably don't want to duplicate that and add our own custom monitors in VS Code.

davidanthoff:
@ZacLN I think this is no longer needed, because you fixed this with another PR? Should we close it?

ZacLN:
I'm not sure, but I'll open an issue and we'll solve it on the TS side if still applicable

PR #159, opened 2017-08-04
directly access string value when linting `include`s

ZacLN:


davidanthoff:
I assume this is for post v0.7.0? I've added it to the LS v0.2.0 milestone, please change if that is incorrect :)

davidanthoff:
Ah, sorry, just realized that this is a bug fix that should go into a RC3, right? I'll create it once you merge this.

ZacLN:
Yeah sorry, just a quick one

PR #163, opened 2017-08-08
don't import `:.`

ZacLN:
Fixes segfaults reported https://github.com/JuliaEditorSupport/julia-vscode/issues/236.
Sorry @davidanthoff, one more fix. Though I don't think this needs a new RC release: it's a one line change cutting out crashes!


davidanthoff:
No worries, these RCs are produced very quickly, and this kind of bug hunting/squashing is great, I think we'll have a much higher quality release this time around than ever before.

PR #165, opened 2017-08-08
allow parameterised type construct names during lint

ZacLN:


mauro3:
Sorry, just a bit too late for 0.7: I tested this and it did not work.  Once I added the additional part of the test to the other two occurrences where the warning "Constructor name does not match" is generated, it worked:

https://github.com/JuliaEditorSupport/LanguageServer.jl/blob/c37b7a7846e13cf20ea6ecc962c7d2c9d1562938/src/trav/lint.jl#L342
https://github.com/JuliaEditorSupport/LanguageServer.jl/blob/c37b7a7846e13cf20ea6ecc962c7d2c9d1562938/src/trav/lint.jl#L365

davidanthoff:
Is this enough of a corner case that we can leave as is for now, or should we do a v0.7.1 bug fix release? Otherwise it would get fixed in v0.8.0, which might be a couple of weeks off...

mauro3:
No need for a bug fix for me, I just updated those two lines in the `~/.vscode/...` folder. Thanks!

Issue #166, opened 2017-08-08
Support for Sublime


davidanthoff:
There seems to be a LSP client for Sublime: https://github.com/sourcegraph/sublime-lsp

Would be good to investigate whether that works with this package, and if not what we need to do to get it working.

randy3k:
There is another alternative, https://github.com/tomv564/LSP

davidanthoff:
That looks good, if someone gets this to work with julia, please report back!

randy3k:
LanguageServer reported `ERROR: KeyError: key "file:///Users/Randy/Dropbox/Julia/BandedMatrices/src/interface.jl" not found` while I tried to run it via LSP. Below is the debug log of LSP which includes the payloads sent and received by LSP and LanguageServer. It might help you to identify the issue.

```
LSP: starting in /Users/Randy/Dropbox/Julia/BandedMatrices
LSP: starting ['/usr/local/bin/julia', '--startup-file=no', '--history-file=no', '-e', 'using LanguageServer; server = LanguageServer.LanguageServerInstance(STDIN, STDOUT, false); server.runlinter = true; run(server);']
LSP: request 1: initialize 
LSP: payload  OrderedDict([('jsonrpc', '2.0'), ('id', 1), ('method', 'initialize'), ('params', {'capabilities': {'textDocument': {'completion': {'completionItem': {'snippetSupport': True}}}}, 'rootUri': 'file:///Users/Randy/Dropbox/Julia/BandedMatrices', 'rootPath': '/Users/Randy/Dropbox/Julia/BandedMatrices', 'processId': 82866})])
LSP: client registered for window 15 {'julia': <LSP.main.Client object at 0x10d2c5490>}
LSP: got json:  {"id":1,"jsonrpc":"2.0","result":{"capabilities":{"textDocumentSync":1,"hoverProvider":true,"completionProvider":{"resolveProvider":false,"triggerCharacters":["."]},"signatureHelpProvider":{"triggerCharacters":["("]},"definitionProvider":true,"referencesProvider":true,"documentHighlightProvider":false,"documentSymbolProvider":true,"workspaceSymbolProvider":true,"codeActionProvider":true,"documentFormattingProvider":true,"documentRangeFormattingProvider":false,"renameProvider":true,"documentLinkProvider":{"resolveProvider":false},"executeCommandProvider":{"commands":[]},"experimental":null}}}
LSP: notify: textDocument/didChange
[...]

PR #177, opened 2017-09-02
script to launch LanguageServer

randy3k:
This is a script which I use in Sublime Text #166  to launch LanguageServer.

The reason of this is because Julia spawns a child process while precompiling packages and that childprocess makes julia so resistant so that julia won't be killed even the editor exits which leaves a orphan julia process. This script detects if the julia process is orphan and kill it and its possible child processes.

This script could be also used in other editors, e.g., emacs/neovim. Or even VSCode, not sure if VSCode is smart enough to kill all child processes.

It should also work on linux, though, I have only tested it on macOS. Also, not sure if Windows suffers from the same issue.

### DIFF
--- a/contrib/languageserver.sh
+++ b/contrib/languageserver.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+JULIABIN="julia"
+DEBUG="false"
+
+while [[ $# -gt 0 ]]
+    do
+    key="$1"
+
+    case $key in
+        -j|--julia)
+            JULIABIN="$2"
+            shift
+        ;;
+        --debug)
+            DEBUG="true"
+        ;;
+        *)
+            echo "unknown option: $key" >&2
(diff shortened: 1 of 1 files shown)


randy3k:
How do you guys think about this?

davidanthoff:
I have to admit I don't really understand it :) I don't have the impression that we need this for VS Code? Or do we? Our whole process management is going to get a bit more complicated going forward and it is not clear to me how this would fit into that story...

Maybe one solution would be to put this into a folder called ``contrib``, with a clear indication that it is not part of the core, maintained part of the package? That would enable folks to use it, play with it etc. while we don't take on a dependency that might make our life more complicated down the road?

randy3k:
@davidanthoff I changed the folder name.

davidanthoff:
Excellent, thanks!

PR #186, opened 2017-09-21
Fix STDERR redirection error

### LINKED ISSUE #185: jl_uv_writecb() ERROR: broken pipe EPIPE
When I open the Query.jl package with the latest master version, I get lots of errors like in the title. [Here](https://gist.github.com/davidanthoff/77db0206023eb111c66ab26a0b986335) is the full debug script.

@ZacLN Do you get the same on Linux, or is this a Windows thing?

davidanthoff:
This fixes #185 on my system.

I don't think we need to redirect ``STDERR`` here, that is not used for the language server communication in any case.

We already redirect ``STDOUT`` [here](https://github.com/JuliaEditorSupport/julia-vscode/blob/master/scripts/languageserver/main.jl#L10), and I think that should be enough to prevent problems in the communication between the LS and vscode.

### DIFF
--- a/src/trav/toplevel.jl
+++ b/src/trav/toplevel.jl
@@ -133,8 +133,6 @@ function import_modules(x::Expr, server)
         if x.args[1] isa Symbol && x.args[1] != :. # julia issue 23173
             topmodname = x.args[1]
             if !isdefined(Main, topmodname)
-                oSTDERR = STDERR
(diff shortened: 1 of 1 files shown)


ZacLN:
We can try this, I think this did the loading of some package but can't remember which now...

PR #190, opened 2017-09-27
add log toggle handler

ZacLN:


ZacLN:
Yep, quite annoyed Ididn't add it before the vsix...

PR #203, opened 2017-10-04
fix module importing

ZacLN:


ZacLN:
Shall we put out a new beta?

PR #208, opened 2017-10-07
add simple macro handling api

ZacLN:
Add's a simple macro handling api. Package macros can now be added to either be ignored by linting or generate toplevel symbols. 
e.g.
```
struct MacroDef
    package::String #currently not used
    contrib_top::Function
    lint::Bool
end
"traitdef" => MacroDef("SimpleTraits", function(x) 
            [str_value(CSTParser.get_id(x.args[2]))]
        end, true)
```
`lint` indicated where to lint the macrocall's arguments and `contrib_top` is a function that determines the names of any variables that the macro adds to the namespace (it should return a vector of strings).  Some understanding of CSTParser is required for the second part.

In the above example applied to `@traitdef MyTrait{G}`, `x.args[2]` defers to the first call argument, `get_id` get's the name of an identifier from an expression containing type declaration parts (i.e. curly braces here) and `str_value` converts an IDENTIFER to a string.


davidanthoff:
It would be nice if one could define these exceptions for a given macro in the package that defines the macro, and not here. Ideally with that package not having to take a dependency on any of our stuff, *and* without us having to load the package into the LS process.

Maybe one way to achieve that would be a convention that inside a package you could do something like:
````julia
const lint_macro_exceptions = [(:name_of_macro, someotherstuff), (:name_of_macro, somemorestuff)]
````
And then our code could detect that statically and use those exceptions?

But fully on board to merge this PR here for now, it improves things greatly and we can figure this other stuff out later.

ZacLN:
Yep I agree, I don't want to be in charge of checking PRs. I just want to have it here for now so that i can adjust the capabilities/api of it until it gets settled down and working

PR #216, opened 2017-11-01
Use URI2 as key for docs

davidanthoff:
The implementation of ``URI2`` now is case-insensitive on Windows, and case-sensitive everywhere else.

This fixes things on Windows, where julia ``include`` statements are case-insensitive.


src/provider_misc.jl
```diff
@@ -311,7 +311,9 @@ end
 
 
 function process(r::JSONRPC.Request{Val{Symbol("julia/toggle-lint")},TextDocumentIdentifier}, server)
-    server.documents[r.uri]._runlinter != server.documents[r.uri]._runlinter
+    # TODO Isn't this always false? I think this is meant to toggle things
+    # i.e. use assignment rather than a comparison here, right?
+    server.documents[filepath_from_uri(r.uri)]._runlinter != server.documents[filepath_from_uri(r.uri)]._runlinter
```
davidanthoff: @ZacLN This is just a bug, right? It should simply flip the value of ``_runlinter``?

ZacLN: yup def a bug

davidanthoff: Alright, I fixed this with a commit, if you could take a brief look at that as well, it would be great. Is part of this PR.

davidanthoff:
Ok, just realized this doesn't work... We can't assume all URIs are file paths, so we'll have to keep using URIs as the key... I have one other idea for this, I'll try that next. Essentially a URI type that compares case-insensitive for files on win.

davidanthoff:
Alright, this should work now and be ready for review.

I'm now using a new type called ``URI`` (to distinguish it from the type in URIParser.jl), that provides the case-insensitive comparison and hashing for ``file://`` URIs on Windows.

PR #229, opened 2017-11-13
handle workspaceFolders

ZacLN:


src/provider_misc.jl
```diff
@@ -425,3 +460,18 @@ end
...
+        load_folder(wksp, server)
+    end
+end
+function JSONRPC.parse_params(::Type{Val{Symbol("workspace/didChangeWorkspaceFolders")}}, params)
+    added = String[]
+    for wksp in params
+        push!(added, wksp["uri"]["fsPath"])
+    end
+    return added
+end
```
davidanthoff: Don't we have to handle ``removed`` here as well?

ZacLN: Once loaded I'm not sure it's worthwhile removing parsed files from the server:
- It isn't a drag on resources
- I'm slightly concerned about someone adding a folder, then a sub folder - then removing one and us removing files that are still loaded

davidanthoff: Hm, isn't there a risk that for example we won't get notifications about changes to such a file anymore from the client, so we'll have a doc loaded in LS but as a stale version? Somehow that sounds like trouble to me, even though I don't know what sort of trouble yet ;)

I guess what we really need is some function that would remove any doc from our doc list that:
1. is not a child of one of the open workspaces
2. and is not open

That doesn't sound too complicated? I kind of feel it might be better to do this right from the get go?

ZacLN: fixed

davidanthoff:
Alright, I'll merge this now and we can fix any remaining bugs with new PRs, seems easier.

PR #232, opened 2017-11-20
fix sig provider with function return decl

ZacLN:


davidanthoff:
This would fix https://github.com/JuliaEditorSupport/julia-vscode/issues/391, right?

ZacLN:
Not sure if its the same issue

PR #235, opened 2017-11-22
Multi root workspace

davidanthoff:
Alright, this seems to mostly work.

I think I still saw an error at some point, but maybe we'll merge this for now and go from there?


davidanthoff:
Test failures fixed

PR #236, opened 2017-11-22
Add ignorelint list

ZacLN:
files or folders can be ignored by adding to : 
```
"julia.lintIgnoreList": [
    ],
```

I'll follow this with some explorer context menus on the client side to allow gui adding of files to the list


sbromberger:
Did this make it into 0.10? I'm getting an "Unknown configuration setting" when trying to add it.

PR #238, opened 2017-12-14
update for cstparser changes

ZacLN:
Updates for changes to CSTParser interface

### DIFF
--- a/src/languageserverinstance.jl
+++ b/src/languageserverinstance.jl
@@ -28,6 +28,8 @@ function send(message, server)
 end
 
 function Base.run(server::LanguageServerInstance)
+    io = open("/home/zac/jls.log", "a")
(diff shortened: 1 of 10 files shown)


src/languageserverinstance.jl
```diff
@@ -29,6 +29,8 @@ function send(message, server)
 end
 
 function Base.run(server::LanguageServerInstance)
+    io = open("/home/zac/jls.log", "a")
```
KristofferC: Absolute path?

ZacLN: I'll pull this change as we've got telemetry and potentially an upstream fix for the self-deleting logs

ZacLN:
Tests pass, there's a backlog on travis that stopped them from completing though

PR #242, opened 2017-12-29
ignore folders without readperm

ZacLN:


src/provider_misc.jl
```diff
@@ -41,6 +41,8 @@ function JSONRPC.parse_params(::Type{Val{Symbol("initialize")}}, params)
     return InitializeParams(params)
 end
 
+hasreadperm(p::String) = div(uperm(p), 0x04) > 0x00
```
davidanthoff: Isn't the canonical way to check for bit flags ``(uperm(p) & 0x04) == 0x04``?

ZacLN: Could well be- I like to re-implement things poorly

davidanthoff: I had to look it up too ;) And it is, and I think it would make it much easier for folks to figure out what is going on :)

PR #245, opened 2018-01-09
work around for inference/error printing error

ZacLN:
Temporary fix for inference error, I'll try and get a mwe and open an issue upstream


rsrock:
This change was causing a stack overflow error when the convert method was called (with >50k stack frames output in the error!). Is there any harm in disabling this locally?  I couldn't find the original issue.  I'm on Julia 0.6.2-pre.50.

KristofferC:
I don't think this should be defined. It is just a Base bug and patching like this will likely lead to unintended side effects.

ZacLN:
Alright, I’ll push a reversion

PR #266, opened 2018-07-06
Remove debug tracing of all communication

davidanthoff:
With https://github.com/JuliaEditorSupport/julia-vscode/pull/522, we don't need our own logging of traffic.


jingpengw:
Does this affect other IDE?  
I am using this package to support spacevim.

PR #273, opened 2018-08-28
julia 0.7/1.0 update

ZacLN:
Updates for julia 0.7/1.0 adding `StaticLint` dependency. Updates language-server-protocol to 3.0 so requires updated julia-vscode version (currently j0.7 branch).


p4n1k0:
Not working in neovim with LanguageClient-neovim.

PR #308, opened 2018-11-14
Fixed loading SymbolServer, as referenced in issue 300

rishisubramanian:
I was having issues precompiling and using the LS since it was complaining about SymbolServer not being found. This fixes the error for me.


davidanthoff:
I think we don't need this, once SymbolServer.jl is registered properly in METADATA.jl. We are just waiting for the grace period of three days to pass, and then this should be sorted out.

davidanthoff:
Uuups, sorry, I completely misunderstood this PR!

davidanthoff:
Thanks!

PR #347, opened 2019-06-08
Handle julia environments

davidanthoff:
@ZacLN Could you take a look at this? In theory it should complete our handling of julia env changes, but in practice it doesn't yet work. As far as I can tell it handles a non-standard environment well at startup, but I think we still need to do something else when we handle the `"julia/activateenvironment"` notification. But I'm not sure what exactly, I don't fully understand the whole innards of the LS enough to fix that. Any pointers or additional commits would be great!


ZacLN:
Yup this looks like it should work from a quick look, I'll investigate

ZacLN:
Switch code from l137 of misc.jl for the following and this seems to work a treat!

```julia
for (uri, doc) in server.documents
        parse_all(doc, server)
    end
```

src/requests/misc.jl
```diff
@@ -116,14 +116,28 @@ function process(r::JSONRPC.Request{Val{Symbol("julia/getCurrentBlockOffsetRange
...
-end
\ No newline at end of file
+    server.env_path = r.params
+    server.symbol_server = StaticLint.SymbolServer.SymbolServerProcess(depot = server.depot_path, environment=server.env_path)
+    @info "Restarted symbol server"
+    server.packages = StaticLint.SymbolServer.getstore(server.symbol_server)
+    @info "StaticLint store set"
+    kill(server.symbol_server)
+
+    if server.workspaceFolders != nothing
```
ZacLN: Switch this for 

```julia
for (uri, doc) in server.documents
        parse_all(doc, server)
    end
```

davidanthoff: Done, thanks!

PR #350, opened 2019-06-12
WIPI - CSTParser 0.6 compat

ZacLN:
Updates for CSTParser/StaticLint.
-  Document source text, CST and attached semantic information is updated incrementally. Semantic info is now persistently attached to CST nodes.
- Better completions (e.g. where a variable is inferred as being a struct we now provide completions for those fields).
- Speed improvements


davidanthoff:
I'm in favor of merging this rather soon than later, so that all the `master` versions of everything work together, even if it is not entirely finished.

PR #352, opened 2019-06-12
Make tests work on any platform

davidanthoff:


ZacLN:
oops! yes that works or alternatively it can be left empty

PR #374, opened 2019-07-22
add format option handlers

ZacLN:
Allows the passing of config options from client to DocumentFormat. Merging should be delayed until tagging of new DocFormat release + https://github.com/julia-vscode/julia-vscode/pull/806 .


non-Jedi:
This broke master for use with `eglot`. The server sends:

```
[ Info: SENT: {"method":"workspace/configuration","id":-100,"params":{"items":[{"scopeUri":null,"section":"julia.format.indent"},{"scopeUri":null,"section":"julia.format.indents"},{"scopeUri":null,"section":"julia.format.ops"},{"scopeUri":null,"section":"julia.format.tuples"},{"scopeUri":null,"section":"julia.format.curly"},{"scopeUri":null,"section":"julia.format.calls"},{"scopeUri":null,"section":"julia.format.iterOps"},{"scopeUri":null,"section":"julia.format.comments"},{"scopeUri":null,"section":"julia.format.docs"},{"scopeUri":null,"section":"julia.format.lineends"}]},"jsonrpc":"2.0"}
```

and eglot responds with

```
[ Info: RECEIVED: {"jsonrpc":"2.0","id":-100,"result":null,"error":null}
```

And of course `length` is not defined for `nothing`.

[Reading the specification](https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/#workspace_configuration), it seems like eglot may be in the wrong here since the type of `result` is supposed to be `any[]`, but I could be misinterpreting it. I'm not very familiar with the language they're using to describe the interfaces.

non-Jedi:
This PR doesn't follow [the spec](https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/#workspace_configuration) in that it sets `scopeUri` to `nothing` when it needs to be a valid `DocumentUri` (a string). Any chance of rolling this PR back and releasing a 1.0.1?

davidanthoff:
Could you try whether https://github.com/julia-vscode/LanguageServer.jl/pull/416 fixes the problem?

If I understand the `@json_read` macro correctly, then it will lead to a situation where serializing a struct with an optional field will actually just not put that field into the JSON representation if the value is `nothing`, which is probably what we want here, right?

non-Jedi:
Ya, that would be what we want. I checked out 7330639 from #416, and I'm still seeing `"scopeUri": nothing`.

PR #376, opened 2019-07-28
documentChanges is nullable

quangIO:
`documentChanges` can be null according to specs. This is needed for other lsp clients, like [lsp4intellij](https://github.com/ballerina-platform/lsp4intellij)


ZacLN:
Yep, seems correct to me. I can't seem to find where the server would receive a `documentChanges` as a request parameter though?

PR #377, opened 2019-07-30
add handler function for `julia/help` request

ZacLN:
handler for https://github.com/julia-vscode/julia-vscode/pull/818

### DIFF
--- a/src/LanguageServer.jl
+++ b/src/LanguageServer.jl
@@ -1,6 +1,7 @@
 module LanguageServer
 using JSON
 using REPL
+using Markdown
 import URIParser
 import DocumentFormat
 import CSTParser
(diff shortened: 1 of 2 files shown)


aviatesk:
The pain point of using LS for this is we can get a documentation that is only available from `Main` module.
What do you think about using our REPL session for this ? I have a working implementation that works for any user package code as well as usual script.

davidanthoff:
I do think we should be able to get the documentation for user code in the LS process because we have a live parsed representation of every file in the workspace here, and that should include any user docs.

But, I don't know how :) Maybe looking at https://github.com/julia-vscode/LanguageServer.jl/blob/master/src/requests/hover.jl might clarify it? Otherwise, @ZacLN  would know :)

ZacLN:
You can pull explicit documentation from the CST easily as in David's link [above](https://github.com/julia-vscode/LanguageServer.jl/blob/master/src/requests/hover.jl#L34)

PR #419, opened 2019-11-15
Interpret Position offset as UTF16

### LINKED ISSUE #401: LanguageServer.jl uses characters for position rather than UTF-16 codeunits
As per the quote in #400:

>A position inside a document (see Position definition below) is expressed as a zero-based line and character offset. The offsets are based on a UTF-16 string representation. So a string of the form a𐐀b the character offset of the character a is 0, the character offset of 𐐀 is 1 and the character offset of b is 3 since 𐐀 is represented using two code units in UTF-16.

The character offset should be in terms of UTF-16 codeunits. As far as
I can tell, LanguageServer.jl only uses UTF-8 internally and works in
terms of characters (codepoints) rather than codeunits. [eglot works around
this](https://github.com/joaotavora/eglot/blob/260f152634df2ba84ef3e51bdfd4f90a20babd9b/eglot.el#L1007)
but not all editors might. I have no idea how VSCode behaves.

So for e.g. the file:

```julia
𐐀𐐀𐐀="hello"
𐐀𐐀𐐀𐐀=𐐀𐐀𐐀
```

Asking for line 1 position 6 should show the hover for `𐐀𐐀𐐀𐐀` since
the 7th UTF-16 codeunit is still within that variable. Instead it
shows the hover for `𐐀𐐀𐐀`:

```
client-request (id:105) Wed Oct 16 16:29:55 2019:
(:jsonrpc "2.0" :id 105 :method "textDocument/hover" :params
          (:textDocument
           (:uri "file:///home/adam/tmp/test.jl")
           :position
           (:line 1 :character 6)))

server-reply (id:105) Wed Oct 16 16:29:55 2019:
(:id 105 :jsonrpc "2.0" :result
     (:contents
      [(:language "julia" :value "𐐀𐐀𐐀 = \"hello\"")]))
```

There's some discussion about the awkwardness of using UTF-16 code units at https://github.com/microsoft/language-server-protocol/issues/376 and a survey of other implementations at https://github.com/Avi-D-coder/lsp-range-unit-survey.

davidanthoff:
Fixes #401. Maybe.

@StefanKarpinski, if you could take a quick look at what I'm doing here, it would be great!

The other question is whether is is really enough to change those two functions, or whether there are other places in the code base that also need to change. I fear that is beyond my understanding of the code base right now, and we'll have to get @ZacLN's input.

@non-Jedi do you have a way to test whether this branch fixes the issue as you have identified it?

### DIFF
--- a/src/document.jl
+++ b/src/document.jl
@@ -50,16 +50,19 @@ end
 """
     get_offset(doc, line, char)
 
-Returns the byte offset position corresponding to a line/character position. 
+Returns the byte offset position corresponding to a line/character position.
 This takes 0 based line/char inputs. Corresponding functions are available for
 Position and Range arguments, the latter returning a UnitRange{Int}.
 """
 function get_offset(doc::Document, line::Integer, character::Integer)
     line_offsets = get_line_offsets(doc)
     offset = line_offsets[line + 1]
-    while character > 0
-        offset = nextind(doc._content, offset)
-        character -= 1
+    for c in Iterators.take(SubString(doc._content, offset), character)
+        if c >= 0x010000
+            offset += 2
+        else
+            offset += 1
+        end
     end
     return offset
 end
@@ -69,9 +72,9 @@ get_offset(doc, r::Range) = get_offset(doc, r.start):get_offset(doc, r.stop)
 
 """
     get_line_offsets(doc::Document)
-    
-Updates the doc._line_offsets field, an n length Array each entry of which 
-gives the byte offset position of the start of each line. This always starts 
+
+Updates the doc._line_offsets field, an n length Array each entry of which
+gives the byte offset position of the start of each line. This always starts
 with 0 for the first line (even if empty).
 """
 function get_line_offsets(doc::Document, force = false)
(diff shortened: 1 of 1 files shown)


non-Jedi:
I don't unfortunately. Position within the file is sent back and forth from and to the language server all over the language server API. The best test at this point is probably to open up a complex file connected to the language server, add a bunch of characters from outside the basic multilingual plane, and then mess around to make sure you don't get offset hovers, incorrect completion, offset codeinfo, and no crashes for attempting out-of-bounds access when the cursor is near the end of lines.

StefanKarpinski:
Looks right to me. Would be good to have some tests, of course.

davidanthoff:
Ok, cool. I'll add tests (and fix the ones that are now broken), and then hopefully @ZacLN can also take a look.

davidanthoff:
While trying to fix the tests I realized that I thoroughly don't understand our policy with respect to 0 and 1 based indices in this code base :)

I would have thought that whole doc offsets are 1 based? But they don't seem to?

I'm afraid @ZacLN will have to carry this over the finishing line, I'm a little lost right now.

davidanthoff:
@ZacLN could you take a look at this? This PR is not correct, I don't understand our internal design what an offset actually is :) But we need something like that because right now we are not interpreting positions correctly... See the issue linked a the top for a bit more background.

I think it would be good if we try to fix this, I'm pretty positive that this explains a number of the failures we are seeing on telemetry...

ZacLN:
Will dig into this tonight

PR #416, opened 2019-11-05
Fix a bug how missing values in ConfigurationItem are handled

davidanthoff:
Fixes https://github.com/julia-vscode/LanguageServer.jl/pull/416.

@non-Jedi does this work now for you?

PR #423, opened 2019-11-30
Fix configuration handling

davidanthoff:
This now always runs the linter, the only thing one can control is whether the messages are actually published to the client. The setting now also works again :)

Now generally fixes how configuration changes are handled, also for format options.

The second linter setting (the ignore list) was not hooked up in any case, so I just removed that.


non-Jedi:
This PR broke the package. Can't import it any more. See CI failure. Also the reason CI is failing on #416 now.

PR #424, opened 2019-12-02
fix for CSTParser 1.1, etc

ZacLN:


src/requests/textdocument.jl
```diff
@@ -256,19 +254,102 @@ function edit_string(text, editrange, edit)
...
+    doc = server.documents[uri]
+    empty!(doc.diagnostics)
+    response =  JSONRPC.Request{Val{Symbol("textDocument/publishDiagnostics")},PublishDiagnosticsParams}(nothing, PublishDiagnosticsParams(doc._uri, Diagnostic[]))
+    send(response, server)
+end 
+
+function clear_diagnostics(server)
+    for (uri, doc) in server.documents
+        clear_diagnostics(uri, server)
+    end
```
davidanthoff: Are the changes around `publish_diangostics` etc on purpose? They undo https://github.com/julia-vscode/LanguageServer.jl/pull/423, which fixed the configuration handling for those cases.

davidanthoff: Also, I removed the type `LSDiagnostic`, but here it is used again, so this probably also just doesn't work as is?

ZacLN: Nope was an error, reverted

ZacLN:
I've bumped the version to 1.1.0, is this appropriate?

Project.toml
```diff
@@ -13,19 +14,19 @@ Tokenize = "0796e94c-ce3b-5d07-9a54-7f471281c624"
...
 [compat]
-CSTParser = "1"
-DocumentFormat = "1.0.1"
+CSTParser = "1.1"
+DocumentFormat = "1.1.1"
 JSON = "0.20, 0.21"
-StaticLint = "1.0.2"
-Tokenize = "0.5.6"
+StaticLint = "1.1"
+Tokenize = "0.5.7"
```
davidanthoff: I'll update them before I tag a new release, but I'd say we merge as is for now and I can fix on `master` before tagging.

PR #433, opened 2019-12-11
fix hovers when overloading imported functions

ZacLN:


src/requests/hover.jl
```diff
@@ -46,7 +46,8 @@ function get_hover(b::StaticLint.Binding, documentation, server)
                 else
                     break
                 end
-                if b.prev isa StaticLint.Binding && b.prev != b && (b.prev.type == getsymbolserver(server)["Core"].vals["Function"] || b.prev.type == getsymbolserver(server)["Core"].vals["DataType"])
+                @info typeof(b.prev)
```
davidanthoff: Remove this `@info`?

ZacLN: Ta

PR #436, opened 2019-12-16
Fix load_rootpath

davidanthoff:
This just rearranges the checks so that it first checks whether the folder exists and there are read permissions, before it calls `isjuliabasedir` (which will crash if the folder doesn't exist.

This should fix a big that came in via crash reporting.


davidanthoff:
@ZacLN are you on slack right now? Might be easier to quickly coordinate the next RC that way?

PR #437, opened 2019-12-17
fix file link clicking

ZacLN:
running on `isfile` on Strings longer than 255 characters is an error


davidanthoff:
Really? That seems very limiting, does that mean we can't support folders with that depth? Is that a limitation of VS Code or the OS or Julia?

ZacLN:
This is on linux and for continuous segments between path separators (i.e. folder names). We could parse each string and check for this but it seems a bit involved

ZacLN:
Sorry, to clarify further:
1. this is for clickable file paths only.
2. we won't try to check whether strings longer than 256 are clickable

Issue #364, opened 2019-07-06
Malformed tooltips for docstrings


AmjadHD:
hovering over identifiers with docstrings, produces malformed tooltips:
![image](https://user-images.githubusercontent.com/33235928/60750297-e2cd3e80-9fa6-11e9-880f-ee2f31157315.png)

this is because, the whole docstring is taken as one value
```
LSP:  --> textDocument/hover
LSP:      {'contents': [{'language': 'julia', 'value': '```\nprint([io::IO], xs...)\n```\n\nWrite to `io` (or to the default output stream [`stdout`](@ref) if `io` is not given) a canonical (un-decorated) text representation of values `xs` if there is one, otherwise call [`show`](@ref). The representation used by `print` includes minimal formatting and tries to avoid Julia-specific details.\n\nPrinting `nothing` is not allowed and throws an error.\n\n# Examples\n\n```jldoctest\njulia> print("Hello World!")\nHello World!\njulia> io = IOBuffer();\n\njulia> print(io, "Hello", \' \', :World!)\n\njulia> String(take!(io))\n"Hello World!"\n```\n'}]}
```
this is the markdown generated (I added \ to before ``` to not highlight them)
```markdown
\```julia
\```
print([io::IO], xs...)
\```

Write to `io` (or to the default output stream [`stdout`](@ref) if `io` is not given) a canonical (un-decorated) text representation of values `xs` if there is one, otherwise call [`show`](@ref). The representation used by `print` includes minimal formatting and tries to avoid Julia-specific details.

Printing `nothing` is not allowed and throws an error.

# Examples

\```jldoctest
julia> print("Hello World!")
Hello World!
julia> io = IOBuffer();

julia> print(io, "Hello", ' ', :World!)

julia> String(take!(io))
"Hello World!"
\```
\```
```
the hover contents should be like this:
```
{
    'contents': [
        '\nprint([io::IO], xs...)'}, \n\nWrite to `io` (or to the default output stream [`stdout`](@ref) if `io` is not given) a canonical (un-decorated) text representation of values `xs` if there is one, otherwise call [`show`](@ref). The representation used by `print` includes minimal formatting and tries to avoid Julia-specific details.\n\nPrinting `nothing` is not allowed and throws an error.\n\n# Examples\n\n', 
        {
            'language': 'jldoctest',
[...]

PR #438, opened 2019-12-17
use MarkupContent for Hovers

ZacLN:
Markedstring is deprecated, this moves us to using MarkupContent for hovers. This removes the need for manual handling of code blocks for SymbolServer documentation (#364).


davidanthoff:
Seems to break our tests.

PR #439, opened 2019-12-17
use StaticLint.CoreTypes

ZacLN:
WIll require tagging of a new release of StaticLint that includes the `CoreTypes` submodule.


davidanthoff:
Ok, ping me when StaticLint is ready to be tagged, I lost track :)

davidanthoff:
Ok, StaticLint v3.0.0 is tagged. You should probably update the lower bound for it in Project.toml here, and then the tests here should go through.

PR #453, opened 2019-12-24
[WIP] Refactor tests suit

dfioravanti:
Hi everyone,

I really like the work that you are doing with this package and I would like to help. While I am studying and trying to understand the codebase, I thought that I could use the occasion to reorganise the test suit and bring it in line with how other packages seams to be doing things. Mostly I am inspired by how [Turing](https://github.com/TuringLang/Turing.jl/tree/master/test) and [DifferentialEquations](https://github.com/JuliaDiffEq/DifferentialEquations.jl/tree/master/test) are doing testing. My idea would be to reorganise the test code as follows
```
test/
    document/
         all_tests.jl
         test1.jl
         test2.jl
    hover/
         all_tests.jl
         test1.jl
         test2.jl
```

Moreover, every file should be possible to run in isolation so that an arbitrary subset of the tests can be run at any time.

Attached you can see the proof of concept changes that I made to the tests for document, it is not done by any stretch of the imagination. If you like I can continue working on this pull request while studying the code. Please let me know if you think that this work is useless or if you have better ideas about how it should be done.


davidanthoff:
I think @ZacLN  should decide this :)

PR #454, opened 2019-12-31
Handle `nothing` received in Language Server Protocol

kdheepak:
There exists keys in the dictionary from parsing the LanguageServerProtocol message that contain the value `nothing`. This can be changed to a `missing`.

### DIFF
--- a/src/protocol/protocol.jl
+++ b/src/protocol/protocol.jl
@@ -49,7 +49,7 @@ macro dict_readable(arg)
                 f = :(dict[$fieldname])
             end
             if field_allows_missing(field)
-                f = :(haskey(dict,$fieldname) ? $f : missing)
+                f = :(haskey(dict,$fieldname) && !isnothing(dict[$fieldname]) ? $f : missing)
             end
             push!(fex.args, f)
         end


davidanthoff:
Why? We represent `null` values in JSON with `nothing`, I don't understand why we would want to convert them to `missing`.

kdheepak:
Because the `struct`s in this package store a `Union{T,Missing}`. As an example, if you want to create an instance of `TextDocumentClientCapabilities`:

https://github.com/julia-vscode/LanguageServer.jl/blob/b9ba36fc96801671752ba15242a3d4fc2e8d25be/src/protocol/initialize.jl#L137-L159

and `ClientCapabilities` is passed a `dict`:

https://github.com/julia-vscode/LanguageServer.jl/blob/b9ba36fc96801671752ba15242a3d4fc2e8d25be/src/protocol/initialize.jl#L183

where one of the values nested inside the dictionary is `Dict{String, Any}("colorProvider" => nothing)`, then you'll get an error when an instance of `struct Capabilities` is being created because `Capabilities` cannot be constructed using `nothing`.

davidanthoff:
So we interpret `missing` as "this field is not present in the JSON", and `nothing` as `null`. So I think maybe the right solution for some of these cases, where a field could either be not present or have a `null` value, would be to add `Missing` to the union? But I'm not sure, I think @ZacLN needs to respond to this.

kdheepak:
Do you mean adding `Nothing` to the `Union`? 

> So we interpret missing as "this field is not present in the JSON", and `nothing` as `null`

This makes sense. I thought this was the case, and initially I wasn't to add `Nothing` to the right `Union`, but I didn't know which all `Union`s to add `Nothing` to.

ZacLN:
Hi, in the example you gave above you're getting an error because `colorCapabilities` can only be either a `Bool` or missing. 

https://github.com/julia-vscode/LanguageServer.jl/blob/b9ba36fc96801671752ba15242a3d4fc2e8d25be/src/protocol/initialize.jl#L174 
provides an example of a field that can be of a specific type, null or missing.

kdheepak:
I revert the changes in this PR and I added a `@info` to `InitializeParams`

```julia
function InitializeParams(dict::Dict)
    @info dict["capabilities"] # <- Add this line
    InitializeParams(Int(dict["processId"]),
    !haskey(dict, "rootPath") ? missing : dict["rootPath"] === nothing ? nothing : DocumentUri(dict["rootPath"]),
    dict["rootUri"] === nothing ? nothing : DocumentUri(dict["rootUri"]),
    if haskey(dict, "initializationOptions") dict["initializationOptions"] else missing end,
    ClientCapabilities(dict["capabilities"]),
    if haskey(dict, "trace") String(dict["trace"]) else missing end,
    !haskey(dict, "workspaceFolders") ? missing : dict["workspaceFolders"] === nothing ? nothing : WorkspaceFolder.(dict["workspaceFolders"]))
end
```

I get the following error:

```
┌ Warning: no Manifest.toml file found, static paths used
└ @ Revise ~/.julia/packages/Revise/S7mrl/src/Revise.jl:1190
[ Info: Started symbol server
[ Info: Tried to load bac558e1-5e72-5ebc-8fee-abe8a469f55d but failed to find it in the manifest.
[ Info: Tried to load 37e2e46d-f89d-539d-b4ee-838fcccc9c8e but failed to find it in the manifest.
[ Info: Tried to load 2f01184e-e22b-5df5-ae63-d93ebab69eaf but failed to find it in the manifest.
[ Info: Tried to load 8f399da3-3557-5675-b5ff-fb832c97cbdb but failed to find it in the manifest.
[ Info: Tried to load b8f27783-ece8-5eb3-8dc8-9495eed66fee but failed to find it in the manifest.
[ Info: Tried to load fdba3010-5040-5b88-9595-932c9decdf73 but failed to find it in the manifest.
[ Info: store set in 0.014111837 seconds
[ Info: Dict{String,Any}("workspace" => Dict{String,Any}("didChangeWatchedFiles" => Dict{String,Any}("dynamicRegistration" => true),"applyEdit" => true),"textDocument" => Dict{String,Any}("definition" => Dict{String,Any}("linkSupport" => true),"declaration" => Dict{String,Any}("linkSupport" => true),"publishDiagnostics" => Dict{String,Any}("relatedInformation" => true),"completion" => Dict{String,Any}("completionItem" => Dict{String,Any}("snippetSupport" => false)),"implementation" => Dict{String,Any}("linkSupport" => true),"signatureHelp" => Dict{String,Any}("signatureInformation" => Dict{String,Any}("parameterInformation" => Dict{String,Any}("labelOffsetSupport" => true))),"typeDefinition" => Dict{String,Any}("linkSupport" => true),"colorProvider" => nothing))
ERROR: MethodError: Cannot `convert` an object of type Nothing to an object of type Bool
Closest candidates are:
  convert(::Type{T}, !Matched::T) where T<:Number at number.jl:6
  convert(::Type{T}, !Matched::Number) where T<:Number at number.jl:7
  convert(::Type{T}, !Matched::Ptr) where T<:Integer at pointer.jl:23
  ...
Stacktrace:
 [1] convert(::Type{Union{Missing, Bool}}, ::Nothing) at ./missing.jl:69
```

Is the client sending this data as `null`? Should I open an issue on the client side then?

ZacLN:
Yes an issue on the client side would be good, we'll have a think on how we want to interact with misspecified messages on our side

davidanthoff:
I think we should throw an error if we receive a message that is not spec conform and quit.

My understanding from the conversation here is that we should close this PR, right? I'm doing that now, if I misunderstood, please let us know!

PR #458, opened 2020-01-02
Remove `import SymbolServer` from languageserver.sh

kdheepak:


ZacLN:
Importing SymbolServer into the same scope as LanguageServer is necessary (for reasons that are unclear) for LanguageServer to work properly

kdheepak:
I'm able to run it without `import SymbolServer`. Also, if this is in the script, users have to manually install `LanguageServer` AND `SymbolServer` manually.

PR #461, opened 2020-01-07
enable action provider, add import vars actions

ZacLN:
As per title, enables CodeActionProviders. Adds the capability of replacing explicit references to package's variables with a `using` statement. e.g.
```julia
using Somepackage
...
Somepackage.somefunc(1,2,3)
...
```
becomes
```julia
using Somepackage
using Somepackage: somefunc
...
somefunc(1,2,3)
...
```


ZacLN:
combined with other related PR

PR #464, opened 2020-01-09
add alternative get_expr for hovers and definitions

ZacLN:
get_expr1 finds the expression at a given offset, choosing an identifier when lying in between two expressions. Fixes https://github.com/julia-vscode/julia-vscode/issues/925 and some other issues where hovers weren't appearing when the cursor was at the start of an expression.


ZacLN:
Erg, I'll dig in to it tomrrow

PR #467, opened 2020-01-11
Handle NONSTDIDENTIFIERs with get_expr

ZacLN:
To interact with https://github.com/julia-vscode/CSTParser.jl/pull/118, may require version changes


davidanthoff:
Ok, CSTParser 2.1.0 is released and tagged, so updating the min version in Project.toml here should get things working.

PR #469, opened 2020-01-13
Async symbolserver

davidanthoff:
Complements https://github.com/julia-vscode/SymbolServer.jl/pull/72. All seems to work :)

EDIT: There is another benefit of this approach: we should be able to eventually just have the JSON RPC layer and the symbol server put their messages into the same channel, and then the main message loop can just process both messages from the client _and_ from the symbol server, without the current design that we only handle symbol server messages when something happened on the main message queue.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -14,9 +14,9 @@ URIParser = "30578b45-9adc-5946-b283-645ec420af67"
 
 [compat]
 CSTParser = "2.1"
-DocumentFormat = "2"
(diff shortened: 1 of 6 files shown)


ZacLN:
Does this needs to have indicators for progress added back in? I've moved everything to master and I can't tell whether everything is being re-cached or its stuck in a loop

davidanthoff:
I’m looking into the progress stuff right now. Need to sort out sending from a background task first, though :)

PR #472, opened 2020-01-13
add sync for client lint options

ZacLN:


ZacLN:
Yep linter now runs and its requires a bump of version for DocumentFormat.

One slight issue - when I was testing this changing "julia.lint...." options didn't trigger a `didConfigurationChange` event, while changing those relating to formatting did. I've got no clue why

davidanthoff:
I tagged DocumentFormat v2.0.0 (needed a major version bump, because the additional option was a breaking change in the API).

You need to add the names of the new options [here](https://github.com/julia-vscode/julia-vscode/blob/master/src/extension.ts#L169) so that they get synced.

ZacLN:
Ah, thanks that makes sense

PR #476, opened 2020-01-20
Send all messages via a queue

davidanthoff:
The idea is that we now never call `write_transport_layer` anywhere, other than in the thread for message sending. Instead, we use `send` everywhere. The benefit is that we can call that from any thread, without messing up the output pipe. So we can for example now send messages back to the client from the thread that runs the async symbol server stuff etc.

Also fixed a number of other bugs that came up.

Once this is merged, I'll open a PR that a) sends crash reports from the symbol server and b) does the progress UI reporting stuff with the new LSP support for that.

### DIFF
--- a/src/languageserverinstance.jl
+++ b/src/languageserverinstance.jl
@@ -44,6 +44,8 @@ mutable struct LanguageServerInstance
     format_options::DocumentFormat.FormatOptions
     lint_options::StaticLint.LintOptions
 
+    out_msg_queue::Channel{Any}
+
     function LanguageServerInstance(pipe_in, pipe_out, debug_mode::Bool = false, env_path = "", depot_path = "")
         new(
             pipe_in,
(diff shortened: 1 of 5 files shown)


src/languageserverinstance.jl
```diff
@@ -94,11 +97,25 @@ end
...
+            rethrow(err)
+        end
+    end
+
     trigger_symbolstore_reload(server)
     
     global T
     while true
         message = read_transport_layer(server.pipe_in, server.debug_mode)
+
```
ZacLN: Do we need to do something more informative here?

ZacLN: That's a really unhelp view of the code, you'll have to click through to the file

davidanthoff: This should only trigger if the input pipe gets closed, and then we just exit the while loop and shutdown. I think whatever we do here doesn't matter much :) I also plan to rewrite the whole read part of our message handling story in a similar way to the write part, so this code won't have a long lifetime.

PR #480, opened 2020-01-22
disable getText request 

ZacLN:


src/requests/textdocument.jl
```diff
@@ -92,8 +92,9 @@ function process(r::JSONRPC.Request{Val{Symbol("textDocument/didChange")},DidCha
     end
     doc = server.documents[URI2(r.params.textDocument.uri)]
     if r.params.textDocument.version < doc._version
-        write_transport_layer(server.pipe_out, JSON.json(Dict("jsonrpc" => "2.0", "method" => "julia/getFullText", "params" => doc._uri)), server.debug_mode)
-        return
+        # write_transport_layer(server.pipe_out, JSON.json(Dict("jsonrpc" => "2.0", "method" => "julia/getFullText", "params" => doc._uri)), server.debug_mode)
+        # return
+        @warn "The client and server have different textDocument versions for $(doc._uri)."
```
davidanthoff: Maybe we should just properly crash in this case? So that we see this situation in crash reporting? I think just calling `error(...)` with as much info as possible would be the right way to do that. We can then still take out for a shipping version.

ZacLN: Yep

PR #484, opened 2020-01-23
Remove LS tracing

davidanthoff:
Replaces #266. I don't think we need this anymore, now that we have the separate trace output channel.

### DIFF
--- a/src/languageserverinstance.jl
+++ b/src/languageserverinstance.jl
@@ -99,7 +99,7 @@ Run the language `server`.
 function Base.run(server::LanguageServerInstance)
     @async for msg in server.out_msg_queue
         try
-            write_transport_layer(server.pipe_out, msg, server.debug_mode)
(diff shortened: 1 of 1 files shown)


ZacLN:
Do we know whether other editors fully support the new approach?

davidanthoff:
No idea :) We can hold off merging this until we hear back from someone with a different editor.

non-Jedi:
Is this removing ability to view the jsonrpc exchanged between languageserver.jl and the client? If so, as far as clients go, I can confirm that both eglot and lps-mode have the ability to view those.

davidanthoff:
Yes, exactly, it wouldn't echo the raw JSON that is sent back and forth anymore.

So I take it that the two emacs based solutions also don't need it, I guess the other client that folks use is vim? Do we have some user who knows that and could check whether it needs it?

PR #485, opened 2020-01-23
Add JSONRPCEndpoint

davidanthoff:
@ZacLN This is a complete rewrite how we handle messages :) Not working yet, just wanted to give you a heads up that it exists.

The basic idea is to get our JSON RPC implementation in order. So this will properly handle responses, correlate them with a request etc. It should greatly simplify the API we have and make things more robust at the same time.

### DIFF
--- a/src/requests/completions.jl
+++ b/src/requests/completions.jl
@@ -10,8 +10,7 @@ end
 JSONRPC.parse_params(::Type{Val{Symbol("textDocument/completion")}}, params) = CompletionParams(params)
 function process(r::JSONRPC.Request{Val{Symbol("textDocument/completion")},CompletionParams}, server)
     if !haskey(server.documents, URI2(r.params.textDocument.uri))
-        send(JSONRPC.Response(r.id, CancelParams(r.id)), server)
-        return
+        error("Received 'textDocument/completion for non-existing document.")
     end
     
     CIs = CompletionItem[]
--- a/src/requests/completions.jl
+++ b/src/requests/completions.jl
@@ -51,7 +51,7 @@ function process(r::JSONRPC.Request{Val{Symbol("textDocument/completion")},Compl
         end
     end
 
-    send(JSONRPC.Response(r.id, CompletionList(true, unique(CIs))), server)
+    return CompletionList(true, unique(CIs))
 end
 
 function get_partial_completion(doc, offset)
--- a/src/requests/features.jl
+++ b/src/requests/features.jl
@@ -22,8 +22,7 @@ end
 JSONRPC.parse_params(::Type{Val{Symbol("textDocument/signatureHelp")}}, params) = TextDocumentPositionParams(params)
 function process(r::JSONRPC.Request{Val{Symbol("textDocument/signatureHelp")},TextDocumentPositionParams}, server)
     if !haskey(server.documents, URI2(r.params.textDocument.uri))
-        send(JSONRPC.Response(r.id, CancelParams(r.id)), server)
-        return
+        error("Received 'textDocument/signatureHelp for non-existing document.")
     end
     doc = server.documents[URI2(r.params.textDocument.uri)] 
     sigs = SignatureInformation[]
@@ -61,7 +60,11 @@ function process(r::JSONRPC.Request{Val{Symbol("textDocument/signatureHelp")},Te
             end
         end
     end
-    (isempty(sigs) || (typof(x) === CSTParser.PUNCTUATION  && kindof(x) === CSTParser.Tokens.RPAREN)) && return send(JSONRPC.Response(r.id, CancelParams(Dict("id" => r.id))), server)
+    if (isempty(sigs) || (typof(x) === CSTParser.PUNCTUATION  && kindof(x) === CSTParser.Tokens.RPAREN))
+        # TODO Is this an error? Or should we return an empty result here?
+        error("Unclear what should happen here.")
+        # return send(JSONRPC.Response(r.id, CancelParams(Dict("id" => r.id))), server)
+    end
 
     if typof(x) === CSTParser.Tokens.LPAREN
         arg = 0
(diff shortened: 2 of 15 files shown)


ZacLN:
Are there reasons why this can't be split to another package (e.g. so it could be used by DebugAdapter...)?

davidanthoff:
> Are there reasons why this can't be split to another package (e.g. so it could be used by DebugAdapter...)?

Yes, that would be the plan. Once it is done and works I'll put it into its own package, for now it is probably easier to get things working when it is in here.

> And is the intention that we can hook the symbolserver process into this and send Notifications to the client (e.g. for the progress tracker)?

I think there are two major benefits: one is that we can better listen for symbol server and client messages at the same time. But I think the biggest benefit is when we send requests from the server to the extension: that code will look like `result = send_request(...)`, and `result` will then hold the response message from the client. And that pattern should work from inside a function that handles a request from the extension to the server. Right now we have to do things like [this](https://github.com/julia-vscode/LanguageServer.jl/blob/master/src/languageserverinstance.jl#L125) to handle these cases, because our main message loop doesn't handle any further incoming messages (including responses) while we process a request. I ran into this with the progress handling: there we need to send a request to the client to create a progress handler, and once we have a response for that we can start sending progress notifications. With the current design we would have to awkwardly split this code over multiple places. I also think that down the road this design might allow us to handle cancel notifications properly (although I don't think that is a high priority).

I think this should also make our implementation more compliant with the spec. I think things like reusing the message id -100 seem to work, but not clear to me that that is actually how we are supposed to do things ;)

davidanthoff:
I just realised there is another benefit: our request handlers now just return the result at the end of the function, instead of manually sending a response message. I think that should make it a lot easier to write tests, right? Because we can now test these functions by just calling them, without any message passing machinery in place.

This still doesn't work, but it is very close, I think I need one more evening and then it might be good. I have a couple of questions about specific cases, I'll post them as comments here.

src/requests/completions.jl
```diff
@@ -10,8 +10,7 @@ end
 JSONRPC.parse_params(::Type{Val{Symbol("textDocument/completion")}}, params) = CompletionParams(params)
 function process(r::JSONRPC.Request{Val{Symbol("textDocument/completion")},CompletionParams}, server)
     if !haskey(server.documents, URI2(r.params.textDocument.uri))
-        send(JSONRPC.Response(r.id, CancelParams(r.id)), server)
-        return
+        error("Received 'textDocument/completion for non-existing document.")
```
davidanthoff: @ZacLN I'm slightly confused by these types of code block, for two reasons: 1) why would we ever be in this state? I would have expected that the extension would always first send a message that a file is opened, before it would ever send a message like this? So I guess it looks to me as if this is more a generic error state that we should never be in, and we should crash, send a report and figure out what is going on? 2) I think the whole `CancelParams` thing is not meant to be used like this? Or at least I can't find the part in the protocol that corresponds to this type of use. I thought `CancelParams` is only used in `$/cancelRequest` notifications, but those are sent not as a response to a request, but from the same party that started a request? I guess I just generally don't understand this part here :)

src/requests/features.jl
```diff
@@ -61,21 +60,24 @@ function process(r::JSONRPC.Request{Val{Symbol("textDocument/signatureHelp")},Te
             end
         end
     end
-    (isempty(sigs) || (typof(x) === CSTParser.PUNCTUATION  && kindof(x) === CSTParser.Tokens.RPAREN)) && return send(JSONRPC.Response(r.id, CancelParams(Dict("id" => r.id))), server)
+    if (isempty(sigs) || (typof(x) === CSTParser.PUNCTUATION  && kindof(x) === CSTParser.Tokens.RPAREN))
+        # TODO Is this an error? Or should we return an empty result here?
+        error("Unclear what should happen here.")
+        # return send(JSONRPC.Response(r.id, CancelParams(Dict("id" => r.id))), server)
```
davidanthoff: Same thing here, it seems to me that sending a `CancelParams` here is not spec conform. Is this an error state? Or should we just send an empty response here?

src/requests/completions.jl
ZacLN: 1) Not sure how the server gets to that state, but in the past it did somehow. May no longer be needed as I think, in general, things are working a bit better - could test removal in the alpha.
2) Yep looks like an error

src/requests/features.jl
ZacLN: Empty response

ZacLN:
> I think that should make it a lot easier to write tests, right? Because we can now test these functions by just calling them, without any message passing machinery in place.

This would be a huge bonus, our tests are a mess...

src/requests/features.jl
davidanthoff: I'm not entirely clear how to resolve that, can you open a PR that fixes this once we have merged this PR here?

src/requests/init.jl
```diff
@@ -108,9 +108,7 @@ function process(r::JSONRPC.Request{Val{Symbol("initialize")},InitializeParams},
         end
     end
 
-    response = JSONRPC.Response(r.id, InitializeResult(serverCapabilities))
-    send(response, server)
-    LanguageServer.send(LanguageServer.JSONRPC.Notification{Val{Symbol("julia/setStatus")},String}("server setting up"), server)
+    return InitializeResult(serverCapabilities)
```
davidanthoff: I'm removing the custom status for now, still hope to get the progress reporting up and running. If we can't figure that out, I'll put this back in.

davidanthoff:
@ZacLN Ready for a review and merge!

I have another idea how we can simplify things a bit more, but this PR is getting big already, so I'll do that in a follow up PR.

I think for now it would be easiest to leave `JSONRPCEndpoints` in here, we can put it into its own package later.

src/requests/features.jl
ZacLN: Will do

davidanthoff:
@ZacLN, oh, one more thing before you review this: I accidentally merged/pushed https://github.com/julia-vscode/SymbolServer.jl/commit/2149d6c3b27ebbdd7e71d7a9b96aa2131e8766d6 without a PR :) Didn't mean to, but accidentally did. I removed all those `@info` calls because output gets very messed up with all the threads running at the same time.

src/requests/init.jl
ZacLN: Shall we leave it in until it's replaced?

ZacLN:
Looks good, I've a couple of related thoughts but will bring them up when we move the JSONRPC stuff out

src/requests/init.jl
davidanthoff: I think it would have to go somewhere else, though, because it should really only be sent after the init sequence is over? Not sure. I'll put it back in tonight if I don't manage to get the progress reporting working this afternoon.

davidanthoff:
@ZacLN can you formally approve the PR? I can't merge otherwise :)

PR #488, opened 2020-01-25
WIP - add parsing/scoping capabilities for edits to single tokens

ZacLN:
An attempt to limit the amount of work done for static analysis. This identifies edits that can be made by altering terminal expr (i.e. those without child expr, such as identifiers and literals) that don't require us to run a full scopepass. 

The (computational) savings should be significant, i.e. it should mean we're doing no unnecessary work when writing numbers, white space, editing identifiers or strings. 

At the moment its delicate, it'll need a bit more work to ensure its robust enough.


davidanthoff:
I've put this on the backlog, or did you want to ship this with the next release?

ZacLN:
Lets wait on this one

PR #497, opened 2020-01-27
Detect whether client supports progress stuff

### LINKED ISSUE #494: Only send progress stuff if client supports it
I need to double check, but presumably the progress reporting is something that not all clients support and we should check before we send it.

davidanthoff:
Fixes #494.

I read the spec again, and I think we do need to check to be spec conform. In particular, we are also sending a `window/workDoneProgress/create` request, and we should only do that if the client supports it.

### DIFF
--- a/src/languageserverinstance.jl
+++ b/src/languageserverinstance.jl
@@ -52,6 +52,8 @@ mutable struct LanguageServerInstance
 
     current_symserver_progress_token::Union{Nothing,String}
 
+    clientcapability_window_workdoneprogress::Bool
+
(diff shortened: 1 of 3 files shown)


ZacLN:
Looks good but could we upgrade our LSP implementation to 3.15 all in one go rather than bit by bit?

davidanthoff:
My vote is to merge as is and then add the rest for v0.14.1? I just don't want us to ship a version of LS that crashes all the other clients, but updating everything to 3.15 seems a fair bit of work, so maybe something we can do later?

Issue #501, opened 2020-01-28
EOFError


davidanthoff:
From crash reporting [here](https://portal.azure.com/#blade/AppInsightsExtension/DetailsV2Blade/ComponentId/%7B%22Name%22%3A%22julia-vscode-dev%22%2C%22SubscriptionId%22%3A%226803c4ed-bcb4-41d4-bceb-7faf5e7a3469%22%2C%22ResourceGroup%22%3A%22Default%22%7D/DataModel/%7B%22eventId%22%3A%22b9bd5d80-4191-11ea-839a-df6149a8f5f6%22%2C%22timestamp%22%3A%222020-01-28T05%3A47%3A37.843Z%22%2C%22cacheId%22%3A%22cb358880-64b0-4836-b24d-c88b426b3d5c%22%2C%22eventTable%22%3A%22exceptions%22%7D).

```
EOFError:
   at read (iobuffer.jl212)
   at read (io.jl697)
   at get_offset(::LanguageServer.Document, ::Int64, ::Int64) (./scripts/languageserver/packages/LanguageServer/src/document.jl68)
   at get_offset (./scripts/languageserver/packages/LanguageServer/src/document.jl84)
   at process(::LanguageServer.JSONRPC.Request{Val{Symbol("textDocument/codeAction")},LanguageServer.CodeActionParams}, ::LanguageServerInstance) (./scripts/languageserver/packages/LanguageServer/src/requests/actions.jl5)
   at run(::LanguageServerInstance) (./scripts/languageserver/packages/LanguageServer/src/languageserverinstance.jl185)
   at top-level scope (./scripts/languageserver/main.jl28)
   at include (boot.jl328)
   at include_relative(::Module, ::String) (loading.jl1105)
   at include(::Module, ::String) (Base.jl31)
   at exec_options(::Base.JLOptions) (client.jl287)
   at _start() (client.jl460)
```

davidanthoff:
There are various explanations for this one, if we exclude incorrect client behaviour, right? Either we are not properly processing some text update from the client and end up with an incorrect document content, or maybe we are missing a line offset computation somewhere?

ZacLN:
I think we should add an incredibly verbose error message to `get_offset`, one that stays just this side of telling us what text was called (i.e. offset position and that searches for the location/frequency of unicode characters). For the beta at least...

ZacLN:
Nothing for us to do before we get some results from this, moving to Next

PR #503, opened 2020-01-28
Add extensive error output for get_offset

davidanthoff:
Might help diagnose https://github.com/julia-vscode/LanguageServer.jl/issues/501.

I tried it once with a fake error, it seems to work. Not the nicest output, but hopefully we'll get what we need. Do you think this is sufficient?

### DIFF
--- a/src/document.jl
+++ b/src/document.jl
@@ -65,7 +65,11 @@ function get_offset(doc::Document, line::Integer, character::Integer)
     io = IOBuffer(get_text(doc))
     seek(io, line_offsets[line + 1])
     while character > 0
-        c = read(io, Char)
+        try
+            c = read(io, Char)
+        catch err
+            error("get_offset crashed. More diagnostics:\nline=$line\ncharacter=$character\nposition(io)=$(position(io))\nline_offsets='$line_offsets'\ntext='$(get_text(doc))'\n\noriginal_error=$(sprint(Base.display_error, err, catch_backtrace()))")
+        end
         character -= 1
         if UInt32(c) >= 0x010000
             character -= 1
--- a/src/document.jl
+++ b/src/document.jl
@@ -63,26 +63,26 @@ function get_offset(doc::Document, line::Integer, character::Integer)
     c = ' '
     line_offsets = get_line_offsets(doc)
     io = IOBuffer(get_text(doc))
-    seek(io, line_offsets[line + 1])
-    while character > 0
-        try
+    try
+        seek(io, line_offsets[line + 1])
+        while character > 0        
             c = read(io, Char)
-        catch err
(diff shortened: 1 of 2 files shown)


ZacLN:
How much overhead is there for wrapping stuff in a try block? This func gets called alot... If the overheads significant (rough guess) then can we wrap the whole function in a try block instead? (i.e. rename the existing `get_offst` -> `_get_offset` and have get_index call _get_index in a try block)

davidanthoff:
Good point :) Here are some benchmarks:

Without try catch
```
julia> @benchmark LanguageServer.get_offset(d, 37, 8)
BenchmarkTools.Trial:
  memory estimate:  64 bytes
  allocs estimate:  1
  --------------
  minimum time:     45.501 ns (0.00% GC)
  median time:      50.960 ns (0.00% GC)
  mean time:        58.058 ns (1.94% GC)
  maximum time:     1.134 μs (93.05% GC)
  --------------
  samples:          10000
  evals/sample:     989
```
With my original implementation:
```
julia> @benchmark LanguageServer.get_offset(d, 37, 8)
BenchmarkTools.Trial:
  memory estimate:  128 bytes
  allocs estimate:  2
  --------------
  minimum time:     167.285 ns (0.00% GC)
  median time:      186.359 ns (0.00% GC)
  mean time:        201.089 ns (1.15% GC)
  maximum time:     1.669 μs (76.86% GC)
  --------------
  samples:          10000
  evals/sample:     755
```
With the new implementation that I just pushed:
```
julia> @benchmark LanguageServer.get_offset(d, 37, 8)
BenchmarkTools.Trial:
  memory estimate:  128 bytes
  allocs estimate:  2
  --------------
  minimum time:     69.928 ns (0.00% GC)
  median time:      79.609 ns (0.00% GC)
  mean time:        95.573 ns (2.44% GC)
  maximum time:     1.142 μs (87.24% GC)
  --------------
  samples:          10000
  evals/sample:     971
```
So still slower, but maybe ok? I think it would be good if we had access to the `position(io)` value, but we only get that if we start the try block after the `IOBuffer` is constructed.

ZacLN:
That looks fine to me (I had no ballpark idea for the slow down)

ZacLN:
Sorry, I missed the part dumping text. Can we obscure the text so that we're not getting sight of users code? Not particularly elegant or efficient but would the following work?
```julia
function obscure_text(s)
    i = 1
    io = IOBuffer()
    while i <= sizeof(s)
        di = nextind(s, i) - i
        if di == 1
            if s[i] in ('\n', '\r')
                write(io, s[i])
            else
                write(io, "a")
            end
        elseif di == 2
            write(io, "α")
        elseif di == 3
            write(io, "—")
        else
            write(io, s[i])
        end
        i += di
    end
    s1 = String(take!(io))
end
```

davidanthoff:
Yes, that would probably work.

I think hunting that bug down will be tricky, we only had three crash reports for it from one user... So I'm not even sure we'll get more reports of this from our beta testers... Can you open a PR with this obscure stuff? I don't have much more time left today :)

ZacLN:
Sure thing!

PR #507, opened 2020-01-30
wait for symserv before sending lint errors

ZacLN:
This stops us sending lint message until we've loaded symbols. It's not clear to me whether the additional field is necessary or we could use one of the existing symserv related fields


ZacLN:
Yeah I think so

davidanthoff:
Ok, I think I don't have cycles for the extension before Sat again (running a workshop the next couple of days), so maybe we'll leave the current beta out there until Sat, hope that someone sends a crash report that helps with the `get_offset` problem, and then shoot for releasing a RC on Sat/Sun?

ZacLN:
Sounds good

PR #529, opened 2020-02-14
check use of sizeof

ZacLN:
Having looked at the jmd code in detail I think we can remove the indexing of strings (as affected by the use of sizeof) and just use the length of the range. Other instances are OK I think except one use in completions.jl that I don't _think_ could ever have an impact


ZacLN:
I've added comments to try to explain my reasoning (or at least provide an evidence trail to where thinking was wrong when things fail in the future...). 

For the parse_jmd change - before we were trying to measure the size of a string by indexing a range, now it just measures the length of the range (which is in bytes, the unit we want here)

PR #530, opened 2020-02-14
check and sync text on save

ZacLN:
The onSave notification now includes the source text as held by the client. This checks whether it matches the server version. It only outputs a warning at the moment, which is of limited use. Is it possible to have it report that this has happened to Azure?


davidanthoff:
Not sure, I'll take a look...

src/requests/textdocument.jl
```diff
@@ -70,6 +70,12 @@ JSONRPC.parse_params(::Type{Val{Symbol("textDocument/didSave")}}, params) = DidS
 function process(r::JSONRPC.Request{Val{Symbol("textDocument/didSave")},DidSaveTextDocumentParams}, server)
     uri = r.params.textDocument.uri
     doc = server.documents[URI2(uri)]
+    if r.params.text isa String
+        if get_text(doc) != r.params.text
+            @warn "Mismatch between server and client text."
+            set_text!(doc, r.params.text)
```
davidanthoff: I would just crash here, this shouldn't happen, right? And then we get a crash report for sure.

PR #532, opened 2020-02-16
Remove trailing whitespace

aminya:
```
git grep -I --name-only -z -e '' | xargs -0 sed -i 's/[ \t]\+\(\r\?\)$/\1/'
```


davidanthoff:
Ok, this makes me too nervous at this point in the release process :) Too many changes in too many places.

I think we should at some point (when it is more stable) just enable the DocumentFormat.jl bot on this repos, that will automatically open PRs whenever there are trailing white spaces (or any other formatting options). The benefit of that is that it does a sanity check to compare the parsed expression trees pre and post formatting, and doesn't format anything if there is a difference.

davidanthoff:
Actually, that formatting job _is_ currently running, but errors for some reason: https://github.com/julia-vscode/LanguageServer.jl/runs/448390017?check_suite_focus=true. I'll have a look why that is happening.

aminya:
That works too

The only repositories that I see have this problems are Julia-VSCode repositories. Maybe there is a problem with the settings of VSCode.
Atom automatically removes trailing whitespaces on save

davidanthoff:
There is a setting in VS Code also to do that, but then it depends on individual contributors to switch that on. The github action should just check `master` after every commit and fix anything, in theory...

aminya:
Agree. But this should be fixed ASAP. Because it makes the history of the PRs ugly. 

For me, For editing each file, Atom fixed all the trailing spaces, and when I wanted to commit, the diff shows a lot of these changes that make the reviewers lost.

davidanthoff:
So I don't think we can fix this ASAP, we need to find a good time for this, ideally after we have shipped 0.14.0 and then also 0.14.1, when we have another window open for more sweeping changes.

Agreed that this is annoying, but I think right now stability of `master` is more important so that we can ship the next two versions. I think your best bet is to turn that setting off for now in Atom when you edit files here...

PR #539, opened 2020-02-19
Fix text sync

davidanthoff:
Alright, here is my attempt at fixing the bugs :) Probably easiest to review commit by commit, I tried to structure them in logical units.

One thing other thing: do we need to call `try_to_load_parents` in `didChange`, and also the file watching notifications? It seems weird to only call that in `didOpen`, right? Although I have to admit that I would almost rather still just get rid of it entirely ;)

### DIFF
--- a/src/document.jl
+++ b/src/document.jl
@@ -52,6 +52,10 @@ function is_workspace_file(doc::Document)
     return doc._workspace_file
 end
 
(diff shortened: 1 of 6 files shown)


davidanthoff:
Closing, some of this is now in https://github.com/julia-vscode/LanguageServer.jl/pull/547, I'll move the other stuff into new PRs.

Issue #544, opened 2020-02-19
Should we handle files outside the workspace at all?


davidanthoff:
My proposal here would be the following:
- we never add a file to the `server._documents` collection if that file is outside of the current workspace, _unless_ we receive a `didOpen` request for such a file. Once such a file outside the workspace is closed again via `didClose`, we remove it from the `server._documents` collection.
- from StaticLint's point of view, the content of `server._documents` is all there exists in terms of code files. If a file is not in that collection, it simply isn't available to StaticLint.
- the code that follows `include` statements would only operate on the files in the `server._documents` collection. If it finds an `include` statement that points to files outside the workspace, it would add a diagnostic message to that particular `include` suggesting that the target folder should be added to the workspace.
- we generally don't publish diagnostics for any file that isn't part of the workspace (I think we should do that no matter what, actually, even if we don't adopt the full idea in this issue here). In particular, we never publish any diagnostic for a file that was opened via `didOpen` but is not part of the workspace.
- the behaviour for files outside the workspace that are opened via `didOpen`, we would offer a reduced functionality: things like code navigation inside the file could still work, but stuff like goto definition not. I think that is actually quite reasonable in any case, because we currently also don't get things right (for example, we don't really know what env we should use to resolve symbols from symbol server for such files).
- we hope that at some point we can utilize [LSIF](https://microsoft.github.io/language-server-protocol/overviews/lsif/overview/) for rich code navigation of files that are not in the workspace. I find the available info on LSIF a bit confusing, and it seems to me that there are some missing pieces in general how this will all work together. But my general sense is that this is the design they are moving towards: LS for files in the workspace, LSIF for stuff from packages etc. So the way this might play out is that we have a LSIF file for Julia base (unclear how we aquire that), and I could imagine a world where the new package server actually hosts LSIF files for packages or something like that. Clearly all future talk, because right now it is quite unclear to me how this would then be integrated into VS Code, but I think generally that seems like a promising direction.

I think the main benefits of this are these:
- Much simplified logic: the LS messages control what we load and when, no "secondary" loading triggers that complicate stuff.
- A much clearer layering. StaticLint gets a read-only view of the world, that is it. I think that will make a multi-threading story _much_ simpler down the road, the kind of design we have right now where StaticLint can trigger new loads makes a simple multi threading design super, super tricky.
- Much clearer defined semantics: currently things like "we look three levels up in the file system" are kind of unpredictable 
[...]

PR #547, opened 2020-02-19
change sequencing for parent-search

ZacLN:


src/requests/textdocument.jl
```diff
@@ -8,22 +8,17 @@ function process(r::JSONRPC.Request{Val{Symbol("textDocument/didOpen")},DidOpenT
         doc._version = r.params.textDocument.version
         get_line_offsets(doc, true)
```
davidanthoff: This branch needs a `set_open_in_editor(doc, true)`.

Also, doesn't this branch also need a call to `try_to_load_parents`?

src/requests/textdocument.jl
```diff
@@ -8,22 +8,17 @@ function process(r::JSONRPC.Request{Val{Symbol("textDocument/didOpen")},DidOpenT
...
-            set_open_in_editor(doc, true)
-            if is_ignored(uri, server)
-                doc._runlinter = false
-            end
+        doc = Document(uri, r.params.textDocument.text, false, server)
+        setdocument!(server, URI2(uri), doc)
+        doc._version = r.params.textDocument.version
+        if any(i->startswith(uri, filepath2uri(i)), server.workspaceFolders)
+            doc._workspace_file = true
         end
```
davidanthoff: `doc._workspace_file = any(i->startswith(uri, filepath2uri(i)), server.workspaceFolders)`

src/requests/textdocument.jl
```diff
@@ -8,22 +8,17 @@ function process(r::JSONRPC.Request{Val{Symbol("textDocument/didOpen")},DidOpenT
...
+        set_open_in_editor(doc, true)
+        if is_ignored(uri, server)
+            doc._runlinter = false
+        end
```
davidanthoff: `doc._runlinter = !is_ignored(uri, server)`

ZacLN:
No, an edit within any given document wont't impact/change whether a parent file that includes it. What's your thinking?

src/requests/textdocument.jl
```diff
@@ -59,10 +52,20 @@ function process(r::JSONRPC.Request{Val{Symbol("textDocument/didClose")},DidClos
     doc = getdocument(server, URI2(uri))
     empty!(doc.diagnostics)
     publish_diagnostics(doc, server)
-    if !is_workspace_file(doc)
-        deletedocument!(server, URI2(uri))
-    else
+    if is_workspace_file(doc)
         set_open_in_editor(doc, false)
+    else
+        if any(d.root == doc.root && doc._open_in_editor for (uri, d::Document) in getdocuments_pair(server) if d != doc)
```
davidanthoff:
```suggestion
        if any(d.root == doc.root for d::Document in getdocuments_value(server) if d != doc)
```
I think we don't need to check for `doc._open_in_editor`, we will only get this message for docs that _are_ open in the editor by definition, right?

src/requests/textdocument.jl
```diff
@@ -59,10 +52,20 @@ function process(r::JSONRPC.Request{Val{Symbol("textDocument/didClose")},DidClos
...
+            # If any other open document shares doc's root we just mark it as closed...
+            set_open_in_editor(doc, false)
+        else
+            # ...otherwise we delete all documents that share root with doc.
+            for (u,d) in getdocuments_pair(server)
+                if d.root == doc.root
```
davidanthoff:
```suggestion
                if d.root == doc.root && !get_open_in_editor(d)
```
We need to make sure that we don't delete files here that are outside the workspace but are currently also open in the editor, I believe? If you accept this change here, you also need to add a `set_open_in_editor(doc, false)` before this loop, otherwise the doc that is actually being close won't get removed.

In general this seems all very tricky and hard to follow logic... if we do this here in `didClose`, do we need to do a similar thing in delete messages from file watch? I think this would all be way, way simpler if we went with https://github.com/julia-vscode/LanguageServer.jl/issues/544 :) I at least have a really hard time figuring out all the possible scenarios/cases here...

src/requests/textdocument.jl
ZacLN: Ooops, this is a typo, should read `d._open_in_editor` which is actually consistent with my comment within the file...

Changing the above addresses your comment below

davidanthoff:
> No, an edit within any given document wont't impact/change whether a parent file that includes it. What's your thinking?

Ah, I didn't get the whole logic here, i.e. that this was looking for other files that are including the file that is being opened. Agreed then.

PR #551, opened 2020-02-19
Throw an error if StaticLint does something it shouldn't do

davidanthoff:
As far as I can tell, StaticLint should never try to set a doc that already exists, this just adds an error if it does so that we would see that in crash reporting. I think if I understand the code correctly, that branch should actually never be hit.

### DIFF
--- a/src/staticlint.jl
+++ b/src/staticlint.jl
@@ -8,7 +8,14 @@ function loadfile(server::LanguageServerInstance, path::String)
     doc = Document(uri, source, true, server)
     StaticLint.setfile(server, path, doc)
 end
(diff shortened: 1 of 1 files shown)


ZacLN:
The only place StaticLint can load a file during running is within `followinclude`, which checks whether a file exists before loading  ([here](https://github.com/julia-vscode/StaticLint.jl/blob/master/src/StaticLint.jl#L129), [here](https://github.com/julia-vscode/StaticLint.jl/blob/master/src/StaticLint.jl#L132), and [here](https://github.com/julia-vscode/StaticLint.jl/blob/master/src/StaticLint.jl#L140)). So I think this PR is OK, if redundant.

davidanthoff:
Yep, hopefully this error will never appear :)

PR #552, opened 2020-02-19
Fix the implementation of didChangeWatchedFiles

davidanthoff:
Some problems with the current implementation:
- it did not properly handle create requests (in particular, if the doc was not already in `server.documents`, it did nothing)
- it did not mark deleted files as no longer in the workspace. That meant that if such a file was open in the editor, and then we received a `didClose`, this file would hang around in `server.documents`, even though it shouldn't.


davidanthoff:
@ZacLN could you review this?

PR #553, opened 2020-02-20
Don't publish diagnostics for out of workspace files

ZacLN:


src/requests/textdocument.jl
```diff
@@ -320,7 +320,7 @@ function mark_errors(doc, out = Diagnostic[])
 end
 
 function publish_diagnostics(doc::Document, server)
-    if server.runlinter && server.symbol_store_ready
+    if server.runlinter && server.symbol_store_ready && doc._workspace_file
```
davidanthoff:
```suggestion
    if server.runlinter && server.symbol_store_ready && is_workspace_file(doc)
```
I think the general convention now is that if a field name starts with `_`, we use accessor functions, right?

ZacLN: Ah right, I hadn't clocked that! Why does this flag need an accessor function?

davidanthoff: I don't know, I thought if we have these for some we should have it for all? Maybe that is silly...

ZacLN: I think the ones for adding/removing documents are needed but for the others the field names are perhaps descriptive enough as they are, I'm not sure whats added  by an accessor func.  We should probably come up with a consistent approach here, but I don't think its _that_ important!

PR #562, opened 2020-02-26
Allow using statements to point to live package code

ZacLN:
Partner of https://github.com/julia-vscode/StaticLint.jl/pull/103

The key question at this point is what files we will allow/determine to be packages. One approach would be to limit it to workspace folders that have the correct structure (i.e. a folder `X` with `src/X.jl` within it)


ZacLN:
https://github.com/julia-vscode/julia-vscode/issues/1023

davidanthoff:
Hm, it seems to me that we generally need to resolve this via the active environment, right? Whereas this just seems to be based on file system structure?

So I think if we see a  `using` statement (that is not a relative path type thingy), then we look into the active env. If the package is deved _and_ the path to the deved package is in our workspace, then we use the live CSTParser tree to resolve symbols, otherwise we use what we get from SymbolServer,  right?

ZacLN:
This is just the internal handling of this (not the env handling part we discuss the other day) and in particular looks to make available the 'main' package when editing it is being edited - this is not information explicitly held in the Project.toml.  I've tightened up the code that decides whether to add a 'live' package to reflect this.

davidanthoff:
I think I still don't fully understand this :)

Lets say someone opens the root folder for package `Foo`. 

Inside the `src` folder we would never expect a `using Foo` statement, right? Only `include` statements to assemble everything together. So I assume that this here is not needed when a user edits the files in `src`, right?

For files outside of `src`, it seems to me that we should _always_ and only go through an environment to resolve `using X` statements. Right now the UI is such that if I open the folder for package `Foo`, and my active env does not have `Foo` deved, I get a warning that I should open an env that contains `Foo`. So the model we have right now is that if someone opens a package folder, then they should have an env active where that package is deved, and then resolving `using Foo` from folders outside of `src` should go via the enviornment, right?

I think right now I just don't understand the scenario where we should try to resolve this purely pased on file location.

ZacLN:
> inside the src folder we would never expect a using Foo statement, right? 

Unfortunately we do, this seems silly. 

Keeping with your example above, if I'm editing Foo I think I want all references to Foo within that context (i.e. the package directory) to point to what I'm editing, not a cached version. If, for example, I'm writing tests I'd like to be able to refer to functions/features I've just added.

ZacLN:
Continuing this - isn't this what we discussed, the cache should only be used for code that is outside the current workspace?

ZacLN:
bump @julia-vscode/core I think this could do with some discussion

src/languageserverinstance.jl
```diff
@@ -40,6 +40,7 @@ mutable struct LanguageServerInstance
     symbol_store::Dict{Symbol,SymbolServer.ModuleStore}
     symbol_extends::Dict{SymbolServer.VarRef,Vector{SymbolServer.VarRef}}
     symbol_store_ready::Bool
+    workspacepackages::Dict{String,Document}
```
davidanthoff, line 43: This should probably have UUIDs as the key, right? I think we are generally not identifying packages by their UUID right now, but probably should to get this right.

src/languageserverinstance.jl
```diff
@@ -114,9 +116,31 @@ end
 
 function setdocument!(server::LanguageServerInstance, uri::URI2, doc::Document)
     server._documents[uri] = doc
+    # Add possible workspace packages
+    path = uri2filepath(uri._uri)
+    for wk_folder in server.workspaceFolders
+        if startswith(path, wk_folder) && normpath(wk_folder) == normpath(server.env_path)
```
davidanthoff, line 122: I don't fully understand the logic here. Is the idea that only if one has activated the package as the env, will any of this work?

src/languageserverinstance.jl
```diff
@@ -114,9 +116,31 @@ end
...
+            sub_path = splitpath(path)
+            first(sub_path) == "/" && popfirst!(sub_path)
+            length(sub_path) < 3 && continue
+            fname = splitext(last(sub_path))[1]
+            if sub_path[end-1] == "src" && sub_path[end-2] == fname
+                @info "Setting $path as source of 'live' package"
+                server.workspacepackages[fname] = doc
```
davidanthoff, line 129: It seems to me that we need to look for a `Project.toml` and then at its content to decide whether something is really a package or not.

src/requests/completions.jl
```diff
@@ -414,6 +421,11 @@ function import_completions(doc, offset, rng, ppt, pt, t, is_at_end ,x, CIs, ser
                     end
                 end
             else
+                for (n,doc1) in server.workspacepackages
+                    if startswith(n, t.val) && StaticLint.has_workspace_package(server, n)
```
davidanthoff, line 425: I think this is where we need to go via the env info: Say the code has `using Foo`, then we need to look in the env what UUID `Foo` corresponds to, then we look in the `server.workspacepackages` that is indexed by UUID whether there is a package with that UUID. But we probably shouldn't match by name, because the names don't have to be the same if the package is loaded via an env (and not `LOAD_PATH`).

davidanthoff (commented):
We clearly need that kind of thing, but I'm wondering whether a) we should really do this pre Juliacon, and b) whether we should maybe first fix our env handling and then do this? Not sure, to be honest.

davidanthoff:
@ZacLN I'd say we can try to get that in soon after Juliacon? Is the next step that you resolve merge conflicts, and then we review properly?

davidanthoff (changes requested):
@ZacLN bump, I think next step is to resolve merge conflicts here.

Issue #411, opened 2019-10-30
server stuck in "Tried to load PACKAGENAME but failed to load from disc, re-caching." loop


orialb:
Hi,

I just tried using the latest master to enjoy some of the recent fixes and I can't get the server to run. 
I first add the latest master in a clean environment and then run the following in the REPL:
```julia
using LanguageServer
server = LanguageServerInstance(stdin, stdout, true, ".", "", Dict())
run(server) 
```

Afterwards the server gets stuck in an infinite loop of trying t load packages from disc:
```julia
[ Info: Started symbol server
[ Info: Tried to load LanguageServer but failed to load from disc, re-caching.
[ Info: Tried to load Dates but failed to load from disc, re-caching.
[ Info: Tried to load StaticLint but failed to load from disc, re-caching.
[ Info: Tried to load Parsers but failed to load from disc, re-caching.
[ Info: Tried to load CSTParser but failed to load from disc, re-caching.
[ Info: Tried to load SHA but failed to load from disc, re-caching.
[ Info: Tried to load Sockets but failed to load from disc, re-caching.
[ Info: Tried to load Random but failed to load from disc, re-caching.
....
```
(this also happens if I use `server = LanguageServerInstance(stdin, stdout, true, "/Users/USERNAME/.julia/environments/v1.1", "", Dict())` )

Julia version:
```julia
julia> versioninfo()
Julia Version 1.1.0
Commit 80516ca202 (2019-01-21 21:24 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin14.5.0)
  CPU: Intel(R) Core(TM) i7-5557U CPU @ 3.10GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, broadwell)
```
(tried also with Julia 1.2) 

Any ideas how to fix / debug this? Is there some cache I should try deleting?

orialb:
I tried to delete the cache in the `store` folder of the relevant `.julia/packages/SymbolServer/` folder, but this didn't help.

orialb:
Never mind, I updated `SymbolServer` to master instead of the version specified in `LangaugeServer` and it works now.

AmjadHD:
Yesterday I downloaded v1.3 of julia and installed LanguageServer.jl v1.0, and this still happens.

AmjadHD:
Anyone seeing this ? @ZacLN ?

[...]

PR #565, opened 2020-02-27
Update example in README.md with SymbolServer

dmolina:
Add using LanguageServer.SymbolServer in the example as it was suggested in #411


davidanthoff:
Do we really need this? I don't really understand why the example in the README would need this.

AmjadHD:
I think `using LanguageServer.SymbolServer` will save the end user one manual install, and what may follow that from dependency mismatch.
but if the readme is updated so should the bash script i guess.

PR #566, opened 2020-02-27
check for invalid char

### LINKED ISSUE #563: Base.InvalidCharError{Char}('\x95')
From crash reporting [here](https://portal.azure.com/#blade/AppInsightsExtension/DetailsV2Blade/ComponentId/%7B%22Name%22%3A%22julia-vscode-dev%22%2C%22SubscriptionId%22%3A%226803c4ed-bcb4-41d4-bceb-7faf5e7a3469%22%2C%22ResourceGroup%22%3A%22Default%22%7D/DataModel/%7B%22eventId%22%3A%2211cd3f03-5979-11ea-b423-455db2d219cb%22%2C%22timestamp%22%3A%222020-02-27T15%3A51%3A35.362Z%22%2C%22cacheId%22%3A%22e15ef07b-b4de-4795-beb5-a4621984e3d2%22%2C%22eventTable%22%3A%22exceptions%22%7D).

Exception type: `Base.InvalidCharError{Char}`
Failed method: `invalid_char(::Char)`
Error message: `Base.InvalidCharError{Char}('\x95') `
Stack trace:
```
Base.InvalidCharError{Char}:
   at invalid_char(::Char) (char.jl85)
   at UInt32(::Char) (char.jl130)
   at mark_errors(::LanguageServer.Document, ::Array{LanguageServer.Diagnostic,1}) (./scripts/languageserver/packages/LanguageServer/src/requests/textdocument.jl291)
   at parse_all(::LanguageServer.Document, ::LanguageServerInstance) (./scripts/languageserver/packages/LanguageServer/src/requests/textdocument.jl264)
   at process(::LanguageServer.JSONRPC.Request{Val{Symbol("workspace/didChangeWatchedFiles")},LanguageServer.DidChangeWatchedFilesParams}, ::LanguageServerInstance) (./scripts/languageserver/packages/LanguageServer/src/requests/workspace.jl41)
   at run(::LanguageServerInstance) (./scripts/languageserver/packages/LanguageServer/src/languageserverinstance.jl217)
   at top-level scope (./scripts/languageserver/main.jl28)
   at include (boot.jl328)
   at include_relative(::Module, ::String) (loading.jl1105)
   at include(::Module, ::String) (Base.jl31)
   at exec_options(::Base.JLOptions) (client.jl287)
   at _start() (client.jl460)
```

ZacLN:
An attempt to fix #563. Assumes that invalid Chars are always `<= 0x010000` (I don't know enough about bitshifting etc to understand `Base.invalid_char`).

### DIFF
--- a/src/document.jl
+++ b/src/document.jl
@@ -77,7 +77,7 @@ function get_offset(doc::Document, line::Integer, character::Integer)
                 character -= 1
             end
         end
-        if UInt32(c) < 0x0080
+        if Base.invalid_char(c) || UInt32(c) < 0x0080
             return position(io)
         elseif UInt32(c) < 0x0800
             return position(io) - 1
--- a/src/document.jl
+++ b/src/document.jl
@@ -73,11 +73,11 @@ function get_offset(doc::Document, line::Integer, character::Integer)
         while character > 0        
             c = read(io, Char)
             character -= 1
-            if !Base.invalid_char(c) && UInt32(c) >= 0x010000
+            if !is_invalid_char(c) && UInt32(c) >= 0x010000
                 character -= 1
             end
         end
-        if Base.invalid_char(c) || UInt32(c) < 0x0080
+        if is_invalid_char(c) || UInt32(c) < 0x0080
             return position(io)
         elseif UInt32(c) < 0x0800
             return position(io) - 1
@@ -174,7 +174,7 @@ function get_position_at(doc::Document, offset::Integer)
     while offset > position(io)
         c = read(io, Char)
         character += 1
-        if !Base.invalid_char(c) && UInt32(c) >= 0x010000
+        if !is_invalid_char(c) && UInt32(c) >= 0x010000
             character += 1
         end
     end
--- a/src/document.jl
+++ b/src/document.jl
@@ -73,7 +73,7 @@ function get_offset(doc::Document, line::Integer, character::Integer)
         while character > 0        
             c = read(io, Char)
             character -= 1
-            if UInt32(c) >= 0x010000
+            if !Base.invalid_char(c) && UInt32(c) >= 0x010000
                 character -= 1
             end
         end
@@ -174,7 +174,7 @@ function get_position_at(doc::Document, offset::Integer)
     while offset > position(io)
         c = read(io, Char)
         character += 1
-        if UInt32(c) >= 0x010000
+        if !Base.invalid_char(c) && UInt32(c) >= 0x010000
             character += 1
         end
     end
--- a/src/requests/textdocument.jl
+++ b/src/requests/textdocument.jl
@@ -288,7 +288,7 @@ function mark_errors(doc, out = Diagnostic[])
             while line_offsets[line] <= offset < line_offsets[line + 1]
                 while offset > position(io)
                     c = read(io, Char)
-                    if !Base.invalid_char(c) && UInt32(c) >= 0x010000
+                    if !is_invalid_char(c) && UInt32(c) >= 0x010000
                         char += 1
                     end
                     char += 1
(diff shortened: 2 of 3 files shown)


davidanthoff:
I guess this is one way, but I almost would prefer if we would try to get to the root of this? So presumably we can't end up in this situation for files that are opened via `didOpen`, because I assume VS Code wouldn't sent us invalid UTF-8. So this is probably a file that we read ourselves from disc?

Maybe what we should do is that we do one pass over the content of a file if we read it ourselves from disc to check for invalid characters, and if there are any, we actually just ignore the file.

So one reason for tackling this earlier is that presumably we have other parts in the code that would also crash if there are invalid characters in the doc, and so fixing this here might just make the LS run a bit longer, to then crash somewhere else?

I guess another explanation for how we might end up with this is that we might have another problem in our doc management, right? It could be that for example we still have some bug in the incremental sync or something like that...

Maybe another idea would be to add a function that checks a whole doc for invalid characters, and then sprinkle calls to that at strategic locations in the code and throw an error if there is a problem (for example, at the end of any `didChange`). If there is some bug in the text sync, that might help us find that?

ZacLN:
I think the issue is that VSCod has no qualms about sending us invalid chars (they are converted to � for display), hence the approach here.

We already have a broad check for valid chars within a file we're going to open: https://github.com/julia-vscode/LanguageServer.jl/blob/master/src/utilities.jl#L120. This is used when loading the workspace at the start and when searching for the parents of an out of workspace file.

We're happy to handle invalid chars _within_ strings in Julia so the currently implemented approach maybe needs to be reconsidered to make it more permissive - but this will require changes to Tokenize/CSTParser

davidanthoff:
> I think the issue is that VSCod has no qualms about sending us invalid chars

I don't think it does. I just tried to open an arbitrary binary file in the editor, and then classified it as a Julia file. We get a `didOpen`, but the actual text we get sent is all valid UTF-8 characters. As far as I can tell, VS Code is doing something to that binary content before it is sending it to us that makes it valid UTF-8.

> This is used when loading the workspace at the start and when searching for the parents of an out of workspace file.

But I think there are _many_ cases where we load files from disc and don't call that function, right? What is not clear to me whether `isvalid(::String)` does the same thing that our `validchars` does?

> We're happy to handle invalid chars within strings in Julia so the currently implemented approach maybe needs to be reconsidered to make it more permissive - but this will require changes to Tokenize/CSTParser

But why would we just not ignore such files? At the end of the day they are buggy files, so I kind of feel just detecting that as early as possible and then just ignoring those files entirely seems by far the easiest approach? That is what I tried to implement in https://github.com/julia-vscode/LanguageServer.jl/pull/567.

ZacLN:
> l text we get sent is all valid..

Ah, that's really useful and is kind of the way I think it should be handled on our side.

> But why would we just not ignore such files? At the end of the day they are buggy files

This is what I am confused about. Files with invalid chars within Char literals, strings and comments will both parse and run. We load files that are invalid in other ways (e.g. syntax) and try to pinpoint where the issues are, I think we should do the same here.

Note - can we split this back to the original [issue](https://github.com/julia-vscode/LanguageServer.jl/issues/563) as this discussion is a bit broader than whats in this PR!

src/document.jl
```diff
@@ -73,11 +73,11 @@ function get_offset(doc::Document, line::Integer, character::Integer)
         while character > 0        
             c = read(io, Char)
             character -= 1
-            if UInt32(c) >= 0x010000
+            if !is_invalid_char(c) && UInt32(c) >= 0x010000
```
davidanthoff: Couldn't we just call `isvalid(c)` here?

ZacLN:
I'm closing this given we're not going to load any files with invalid chars. Let me know if I've misunderstood

PR #567, opened 2020-02-28
Validate utf8

davidanthoff:
This is an alternative "fix" for #566, i.e. it is an alternative for https://github.com/julia-vscode/LanguageServer.jl/pull/566.

This PR here adds really two things:
- Whenever we load files from disc, it checks whether they are valid UTF-8 strings, if not, they are not loaded. We should probably have that no matter what, given that users can always have some invalid file laying around.
- It errors if we receive an invalid UTF-8 string from the language server client. It also checks for valid UTF-8 after we have applied incremental sync changes, so it would crash if there is still some problem there.

This PR might just fix the problem (if the source is an invalid file we read), or it might help track down a problem we still might have with text sync.

### DIFF
--- a/src/requests/init.jl
+++ b/src/requests/init.jl
@@ -84,6 +84,7 @@ function load_folder(path::String, server)
                         continue
                     else
                         content = read(filepath, String)
+                        isvalid(content) || continue
                         doc = Document(uri, content, true, server)
                         setdocument!(server, URI2(uri), doc)
                         parse_all(doc, server)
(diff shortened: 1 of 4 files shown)


davidanthoff:
Maybe it would be a better idea to actually add the invalid docs still to our documents list, but with no actual text and a flag that they are invalid, and then publish a diagnostic? But maybe that is something for the next version ;)

ZacLN:
We'd certainly need to add the ghost documents otherwise we'd crash everytime we got a request related to them. I also things we'd need to ensure that the doc's had a CST that match the invalid text in terms of length

davidanthoff:
> We'd certainly need to add the ghost documents otherwise we'd crash everytime we got a request related to them.

I don't think we can get any requests other than the file system events for them, right? But I think you are right, having them as ghosts seems a better idea.

ZacLN:
Wouldn't we get all the usual requests as the client is handling a (cleaned so as to be valid) copy?

davidanthoff:
I think we only get all the other requests if the doc was opened via `didOpen` before? And those we would have in our list, right?

davidanthoff:
But let me change this PR here a bit so that we have a kind of ghost doc here.

davidanthoff:
Closing this for now, if we do go with something like that I'd reimplement it on top of https://github.com/julia-vscode/LanguageServer.jl/pull/574.

Issue #564, opened 2020-02-27
Mismatch between server and client text


davidanthoff:
:( still not fixed...

From [crash reporting](https://portal.azure.com/#blade/AppInsightsExtension/DetailsV2Blade/ComponentId/%7B%22Name%22%3A%22julia-vscode-dev%22%2C%22SubscriptionId%22%3A%226803c4ed-bcb4-41d4-bceb-7faf5e7a3469%22%2C%22ResourceGroup%22%3A%22Default%22%7D/DataModel/%7B%22eventId%22%3A%22ab468215-5972-11ea-bd42-e7b4475e4e65%22%2C%22timestamp%22%3A%222020-02-27T15%3A05%3A46.973Z%22%2C%22cacheId%22%3A%22e4af13e1-ae64-4f19-adac-89dbc3d04702%22%2C%22eventTable%22%3A%22exceptions%22%7D).

Error message: `Mismatch between server and client text for file:///XXX.jl.`

Stack trace:
```
ErrorException:
   at error(::String) (error.jl33)
   at process(::LanguageServer.JSONRPC.Request{Val{Symbol("textDocument/didSave")},LanguageServer.DidSaveTextDocumentParams}, ::LanguageServerInstance) (./scripts/languageserver/packages/LanguageServer/src/requests/textdocument.jl79)
   at run(::LanguageServerInstance) (./scripts/languageserver/packages/LanguageServer/src/languageserverinstance.jl217)
   at top-level scope (./scripts/languageserver/main.jl28)
   at include (boot.jl328)
   at include_relative(::Module, ::String) (loading.jl1105)
   at include(::Module, ::String) (Base.jl31)
   at exec_options(::Base.JLOptions) (client.jl287)
   at _start() (client.jl460)
```

davidanthoff:
I'm pretty lost about this one... Maybe we need to add some more diagnostics in this crash situation, for example whether that file was open in the editor or not, or something like that? Or maybe the version number, maybe that could tell us something interesting?

We could also turn off the incremental sync for a while and see whether we still get these kind of errors? That would maybe allow us to tell whether there is some problem in that?

davidanthoff:
There are now many more crash reports like that. The ones I looked at all had a pretty high version field (we now get a bit more diagnostics), and they are all open in the editor... My best guess is that there is simply some error in the way we handle the `didChange` request?

One thing we could try is to temporarily disable incremental sync and check whether that gets rid of the error?

[...]

PR #568, opened 2020-02-28
Add more diagnostics for version mismatch case

davidanthoff:
Might help us pin down what is going on in https://github.com/julia-vscode/LanguageServer.jl/issues/564.


davidanthoff:
@ZacLN do you think we can merge this soon? I'd like to push out a new insider build with this, it might help us pin down why we still have a text sync error.

PR #576, opened 2020-03-03
Make deletedocument! more robust

davidanthoff:


src/languageserverinstance.jl
```diff
@@ -117,7 +117,17 @@ function setdocument!(server::LanguageServerInstance, uri::URI2, doc::Document)
...
 function deletedocument!(server::LanguageServerInstance, uri::URI2)
+    doc = getdocument(uri)
+
     delete!(server._documents, uri)
+
+    for d in getdocuments_value(server)
+        if d.root===doc
+            d.root = nothing
+            # TODO Check whether we need to do something else here as well,
+            # maybe rerun the linter or something like that?
```
davidanthoff, line 128: @ZacLN are we sure we don't need to do antyhing like that here?

ZacLN: Sorry, I missed the comment. Yes we do, we want to run a scope pass on these files. `scopepass(d, d)` No other special handling should be required

davidanthoff:
I added a call to `scopepass(d, d)`, but it seems a bit weird to me. Every other call to `scopepass` (as far as I can tell) passes the results from a call to `getroot` as the first argument, so this seems to be an exception? And it seems the only code pass where this can happen?

davidanthoff:
@ZacLN any thoughts on my question re `scopepass` from above?

ZacLN:
It's the exception because we've just deleted the root. We assume these files are now their own roots (this will be modified when included by another file during a scopepass).

davidanthoff:
But can't that situation arise in a freshly loaded workspace as well? If there is just a stray file that is not inlcuded anywhere? And wouldn't that essentially be handleded as `scopepass(nothing, d)`?

ZacLN:
The root is where we start the pass, the 2nd arg is an indicator of what file (if any) we're particularly interested in (so that we can, if we want, give more attention to it). The target file isn't currently used for anything interesting

davidanthoff:
How do stray files then ever get a scope pass in "normal" situations? Say we load a workspace, there is a stray file that is not inlcuded by anthing. Wouldn't that doc have `root==nothing`, and then I just can't find the place in code where this file would get a scopepass. Maybe I'm just looking for the wrong thing?

ZacLN:
There only two call sites for `setroot` 
 - on [creating](https://github.com/julia-vscode/LanguageServer.jl/blob/master/src/document.jl#L23) a Document it is set as it's own root 
- and on successfully following an `include` [call](https://github.com/julia-vscode/StaticLint.jl/blob/master/src/StaticLint.jl#L156) during a scopepass

I think the typing of the field as a union with Nothing because I wasn't aware you could initialize a struct that self referenced from the inner constructor when this was written (I think you can, right?)

davidanthoff:
So maybe then the code here should be:
```julia
d.root = d
scopepass(getroot(d), d)
```
instead? Might amount to the same thing, but it probably is easier for folks to understand how that relates to the normla behavior?

ZacLN:
Thats fine

PR #579, opened 2020-03-05
Add text sync diagnostics

davidanthoff:
Requires https://github.com/julia-vscode/julia-vscode/pull/1034.

As is not very useful, but the mechanics of getting the full text doc seem to work. I double checked, and if one types quickly, there are a fair number of requests where the version in the client is already ahead of the request, so we won't get a version all the time. But still, fingers crossed, we'll get something useful sometimes.

The real question is what extra fields we should include in the exception type, i.e. what kind of diagnostics are useful to receive in the crash report... Maybe the lenght of the doc, the offset positions, and then positions and length of each text edit?

### DIFF
--- a/src/exception_types.jl
+++ b/src/exception_types.jl
@@ -5,3 +5,11 @@ end
 function Base.showerror(io::IO, ex::LSSymbolServerFailure)
     print(io, ex.msg)
 end
+
+struct LSTextSyncError <: Exception
+    msg::AbstractString
+end
+
+function Base.showerror(io::IO, ex::LSTextSyncError)
+    print(io, ex.msg)
+end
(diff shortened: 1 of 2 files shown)


ZacLN:
How about we add a Dict field to LanguageServerInstance (keys are uris, values an array) and store the last 10 results of getText for each Document? That way we increase our changes of being able to make the comparison. Happy to make a pull against this to that effect. (the PR against vscode would need to be changed to alwayrs return the text + version number)

davidanthoff:
I think we would need to put that dict in the extension, right? The problem is that when we receive the `didChange` request in the server, the client has already moved on, so we can no longer get that version of the doc from the client. But if we stored the last 10 versions in the extension, that we should be able to respond to all all these requests from the server, right?

ZacLN:
I'd rather avoid trying to track this in both the client and server (but only because I'm lazy). #581 could work?

PR #581, opened 2020-03-06
Text sync diagnostic2

ZacLN:
Extension of #579. 
Attempts to handle the server lagging behind the server. 

* When the client is ahead of the server it stores the the content changes (and before and after content). 
* When the server text and version matches that received from the client we clear the store.
* When the versions match between client and server but the text doesn't we dump the store (obscuring user text). (I'll then have to reconstruct what has happened in a fairly painful manner!)

This requires a change to https://github.com/julia-vscode/julia-vscode/pull/1034#event-3102414589 so that we always receive the latest version from the client, regardless of the version

### DIFF
--- a/src/exception_types.jl
+++ b/src/exception_types.jl
@@ -5,3 +5,11 @@ end
 function Base.showerror(io::IO, ex::LSSymbolServerFailure)
     print(io, ex.msg)
 end
+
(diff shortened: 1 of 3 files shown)


davidanthoff:
I think we can close this, right?

PR #586, opened 2020-03-08
Don't crash on delete msg for non existing file

davidanthoff:
Fixes a bug from crash reporting.


src/requests/workspace.jl
```diff
@@ -50,19 +50,21 @@ function process(r::JSONRPC.Request{Val{Symbol("workspace/didChangeWatchedFiles"
                 parse_all(doc, server)
             end
         elseif change.type == FileChangeTypes["Deleted"]
-            doc = getdocument(server, URI2(uri))
+            if hasdocument(server, URI2(uri))
```
davidanthoff: @ZacLN this diff for some reason looks more involved than it should. The only real change here is that I added this `hasdocument` check.

PR #588, opened 2020-03-09
Fix UTF-8 validation and file IO error checking

davidanthoff:
When I added the more specific error handlers, I forgot to also adjust the validation phase accordingly. Fixes a bug from crash reporting.

I added some stuff, namely it is now also handling `SystemError` properly, because those can also happen from file IO problems.

### DIFF
--- a/src/requests/textdocument.jl
+++ b/src/requests/textdocument.jl
@@ -402,10 +402,7 @@ function search_for_parent(dir::String, file::String, drop = 3, parents = String
             # Could be sped up?            
             content = try
                 s = read(filename, String)
-                # We throw an error in the case of an invalid
-                # UTF-8 sequence so that the same code path
-                # is used that handles file IO problems
-                isvalid(s) || error()
+                isvalid(s) || continue
                 s
             catch err
                 isa(err, Base.IOError) || rethrow()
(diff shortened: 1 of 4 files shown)


src/requests/textdocument.jl
```diff
@@ -427,10 +424,7 @@ function is_parentof(parent_path, child_path, server)
...
         content = try
             s = read(parent_path, String)
-            # We throw an error in the case of an invalid
-            # UTF-8 sequence so that the same code path
-            # is used that handles file IO problems
-            isvalid(s) || error()
+            isvalid(s) || return false
             s
         catch err
             isa(err, Base.IOError) || rethrow()
```
ZacLN: As per my other comment

ZacLN:
Not quite sure if those comments are clear: could we avoid rethrowing the error on lines 408 and 427 of textdocument.jl

davidanthoff:
> Not quite sure if those comments are clear: could we avoid rethrowing the error on lines 408 and 427 of textdocument.jl

So that is the idea: it will not rethrow for `IOError` (and now `SystemError`), but for any other error it will rethrow (under the assumption that that would indicate a problem unrelated to file IO which we want to see in the crash reports).

ZacLN:
Ah got you! ignore me..

davidanthoff:
:) can you approve?

PR #590, opened 2020-03-10
Guard against more IO errors

davidanthoff:
Fixes a bug from crash reporting, and I also added more checks around a couple other file IO operations. `isdir`, `ispath`, and `readdir` can all crash if there is an IO problem.

### DIFF
--- a/src/requests/completions.jl
+++ b/src/requests/completions.jl
@@ -1,13 +1,3 @@
-_ispath(s) = false
-function _ispath(s::String)
-    try
-        return ispath(s)
-    catch err
-        isa(err, Base.IOError) || isa(err, Base.SystemError) || rethrow()
-        return false
-    end
-end
-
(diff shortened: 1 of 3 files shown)


src/requests/completions.jl
```diff
@@ -302,7 +292,7 @@ function path_completion(doc, offset, rng, t, CIs)
     if !startswith(path, "/")
         path = joinpath(_dirname(uri2filepath(doc._uri)), path)
     end
-    if _ispath(path)
+    try
```
ZacLN: Does this need to be around the whole loop?

davidanthoff: I tweaked it a bit. I think the try catch around the whole loop should stay because `readdir` can fail (and in that case we don't want to run the loop, so having the loop inside the `try` seems good on that). But I added another `try catch` inside the loop, so that if we get an IO error for one of the paths it will just ignore that single path, not the entire directory.

davidanthoff: And now we need a new approve from you because of the push :)

Issue #594, opened 2020-03-13
ArgumentError: embedded NULs are not allowed in C strings


davidanthoff:
From [crash reporting](https://portal.azure.com/#blade/AppInsightsExtension/DetailsV2Blade/ComponentId/%7B%22Name%22%3A%22julia-vscode%22%2C%22SubscriptionId%22%3A%226803c4ed-bcb4-41d4-bceb-7faf5e7a3469%22%2C%22ResourceGroup%22%3A%22Default%22%7D/DataModel/%7B%22eventId%22%3A%22413cf7b2-6538-11ea-bdc4-5bf5a63cfddc%22%2C%22timestamp%22%3A%222020-03-13T14%3A37%3A50.930Z%22%2C%22cacheId%22%3A%2220f3b128-2132-4027-82f4-d3de0fe22123%22%2C%22eventTable%22%3A%22exceptions%22%7D).

More info in the link.

Stack trace:
```
ArgumentError:
   at unsafe_convert (c.jl216)
   at stat(::String) (stat.jl67)
   at isfile (stat.jl311)
   at process(::LanguageServer.JSONRPC.Request{Val{Symbol("textDocument/definition")},LanguageServer.TextDocumentPositionParams}, ::LanguageServerInstance) (.\scripts\languageserver\packages\LanguageServer\src\requests\features.jl116)
   at run(::LanguageServerInstance) (.\scripts\languageserver\packages\LanguageServer\src\languageserverinstance.jl233)
   at top-level scope (.\scripts\languageserver\main.jl28)
   at include (boot.jl328)
   at include_relative(::Module, ::String) (loading.jl1105)
   at include(::Module, ::String) (Base.jl31)
   at exec_options(::Base.JLOptions) (client.jl287)
   at _start() (client.jl460)

```

davidanthoff:
[...]

PR #595, opened 2020-03-13
Remove textDocument/definition for paths in strings

davidanthoff:
Fixes https://github.com/julia-vscode/LanguageServer.jl/issues/594, by removing the feature entirely :)

I think we shouldn't have this irrespective of the bug, the semantics of goto definition are pretty clear in my mind, and following strings if they are filenames doesn't seem to fall under that. I also checked this, and for example the typescript extension also doesn't do that.

### DIFF
--- a/src/requests/features.jl
+++ b/src/requests/features.jl
@@ -110,18 +110,6 @@ function process(r::JSONRPC.Request{Val{Symbol("textDocument/definition")},TextD
                 b = nothing
             end
         end
-    elseif x isa EXPR && typof(x) === CSTParser.LITERAL && (kindof(x) === Tokens.STRING || kindof(x) === Tokens.TRIPLE_STRING)
-        if sizeof(valof(x)) < 256 # AUDIT: OK
-            try
-                if isfile(valof(x))
-                    push!(locations, Location(filepath2uri(valof(x)), Range(0, 0, 0, 0)))
-                elseif isfile(joinpath(dirname(uri2filepath(doc._uri)), valof(x)))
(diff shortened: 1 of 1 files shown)


davidanthoff:
@ZacLN are you ok with this one?

ZacLN:
I don't think so - goto for a reference to a file seems like a reasonable and useful implementation and not something to jettison because of a crash report

davidanthoff:
Wouldn't it make more sense to implement this as [document links](https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/#textDocument_documentLink)?

It just seems weird to implement this as a goto definition response. The spec specifically describes the goto definition as `resolve the definition location of a symbol`, which I think this doesn't fall under.

ZacLN:
Yeah that's fine, its just means the server is doing more work to provide the same feature

davidanthoff:
On the positive side, though, we'd probably have a nicer UI? I assume it would show more like a link, so might be more discoverable?

How do we proceed here? I think the number one priority should be to fix the bug. So I'd say any of the following are fine:
1) we come up with a pure bug fix PR and leave the functionality as is for now
2) we merge this PR here and later add either the functionality here back in or add the link stuff

ZacLN:
I think 1) for now (I think the functionality should remain regardless, so the bug needs to be fixed), then decide whether we want to add additional documentLink functionality. If we do implement documentLink for other purposes then we should def move this code to be handled there, but I don't have an idea at the moment of what other DocumentLinks we'd provide.

davidanthoff:
@ZacLN I'm closing this PR for now and assigned the [original bug report](https://github.com/julia-vscode/LanguageServer.jl/issues/594) back to you :)

PR #614, opened 2020-04-13
update for SS changes

ZacLN:
https://github.com/julia-vscode/SymbolServer.jl/pull/123/

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -17,11 +17,11 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 CSTParser = "2.2"
 DocumentFormat = "2.1"
 JSON = "0.20, 0.21"
-StaticLint = "3.1"
+StaticLint = "4"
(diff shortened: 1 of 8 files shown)


davidanthoff:
StaticLint.jl v4.0.0 is also tagged, so updates to Project.toml should make this PR pass tests :)

Once this one is merged, I would release a new version to the insiders channel and unless we see regressions there for say 24 hours we release it to everyone? I think if you have some cycles right now it would be better we push this out quickly and get good feedback from the larger user base and then just fix any problems quickly. I kind of like our current rhythm of shipping new versions early and often :)

ZacLN:
Yep that will work. This is almost certain to cause some breakages I haven't anticipated but will be able to jump on them quickly

davidanthoff:
Hm, still some problems with tests?

davidanthoff:
This change is not breaking, correct? So I can release a new version with a minor version bump?

ZacLN:
yep

Issue #648, opened 2020-04-26
BoundsError: attempt to access 0-element Array{Any,1} at index [1] 


davidanthoff:
From [crash reporting](https://portal.azure.com/#blade/AppInsightsExtension/DetailsV2Blade/ComponentId/%7B%22Name%22%3A%22julia-vscode%22%2C%22SubscriptionId%22%3A%226803c4ed-bcb4-41d4-bceb-7faf5e7a3469%22%2C%22ResourceGroup%22%3A%22Default%22%7D/DataModel/%7B%22eventId%22%3A%22874cc961-873e-11ea-84c3-bf4f2fa54ef2%22%2C%22timestamp%22%3A%222020-04-25T21%3A48%3A26.456Z%22%2C%22cacheId%22%3A%22fcd2e963-df13-4965-be22-688695f38c3f%22%2C%22eventTable%22%3A%22exceptions%22%7D).

Error message:
```
BoundsError: attempt to access 0-element Array{Any,1} at index [1] 
```

Stack trace:
```
BoundsError:
   at getindex (array.jl729)
   at show_import_path(::Base.GenericIOBuffer{Array{UInt8,1}}, ::Expr) (show.jl1082)
   at show_import_path(::Base.GenericIOBuffer{Array{UInt8,1}}, ::Expr) (show.jl1079)
   at show_unquoted(::Base.GenericIOBuffer{Array{UInt8,1}}, ::Expr, ::Int64, ::Int64) (show.jl1438)
   at show_unquoted_quote_expr(::Base.GenericIOBuffer{Array{UInt8,1}}, ::Any, ::Int64, ::Int64) (show.jl1042)
   at show(::Base.GenericIOBuffer{Array{UInt8,1}}, ::Expr) (show.jl750)
   at show_unquoted(::Base.GenericIOBuffer{Array{UInt8,1}}, ::Expr, ::Int64, ::Int64) (show.jl1455)
   at show_block(::Base.GenericIOBuffer{Array{UInt8,1}}, ::Symbol, ::Array{Any,1}, ::Expr, ::Int64) (show.jl920)
   at show_block(::Base.GenericIOBuffer{Array{UInt8,1}}, ::Symbol, ::Symbol, ::Expr, ::Int64) (show.jl929)
   at show_unquoted(::Base.GenericIOBuffer{Array{UInt8,1}}, ::Expr, ::Int64, ::Int64) (show.jl1284)
[...]

PR #650, opened 2020-04-27
wrap hover fallback in try block

ZacLN:
#648


davidanthoff:
I merged right away to get this into a new build!

PR #664, opened 2020-05-05
Remove debug_mode field from LanguageServerInstance

### LINKED ISSUE #421: Log debug messages with @debug instead of with server.debug_mode && @info
The `debug_mode` field of `LanguageServerInstance` should be eliminated. Instead, all debug output that's currently gated with `server.debug_mode && @info "debug log"` should simply be `@debug "debug log"`. This way, logging can be controlled with the standard tools from the `Logging` library instead of having to be aware of the `debug_mode` field.

I think this is better than eliminating logging as in #266 as it can be useful for debugging other clients even if unneeded for vscode.

Obviously, this is a rather simple set of PRs to LanguageServer and SymbolServer, but I wanted to make sure there was support for this before doing the work. This would also be a breaking change and so would probably require a version bump to 2.0...

non-Jedi:
There are no longer any uses of this field inside
LanguageServer.jl. If the vs-code extension still sends
"julia/toggle-log" methods, it should be removed since it functionally
changes nothing currently.

Closes #421

I couldn't get the tests to run locally, so I'm depending on CI to see if anything is broken.


non-Jedi:
Note that this requires a major version bump since the signature of the single documented, exported symbol changed.

davidanthoff:
@non-Jedi, can you resolve the merge conflict?

non-Jedi:
Rebased and force-pushed to resolve merge conflict.

PR #666, opened 2020-05-06
Pass nothing directly to config option constructors

### LINKED ISSUE #624: Default workspace/configuration to same values as vscode extension
Since in the absence of a configuration value, all configuration options default to off, most non-vscode users probably aren't taking advantage of linting and formatting (or aren't aware such capabilities exist). Instead of defaulting to off, would you accept a PR that changes the default values in `request_julia_config` to [match the vs-code defaults](https://github.com/julia-vscode/julia-vscode/blob/1c50151bf7951cdf522a118bd70df2bd51425a21/package.json#L343)?

non-Jedi:
With the PRs below, this causes options to default to the value
specified as a default by the DocumentFormat and StaticLint packages.

Requires
- https://github.com/julia-vscode/DocumentFormat.jl/pull/94
- https://github.com/julia-vscode/StaticLint.jl/pull/153

Fixes #624

----

#

### DIFF
--- a/src/requests/workspace.jl
+++ b/src/requests/workspace.jl
@@ -81,35 +81,13 @@ function request_julia_config(server)
         ]))
     
     # TODO Make sure update_julia_config can deal with the response
-    if length(response) == length(fieldnames(DocumentFormat.FormatOptions)) + 1 + length(fieldnames(StaticLint.LintOptions))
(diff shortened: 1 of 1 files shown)


non-Jedi:
obsoleted by #683

PR #678, opened 2020-05-07
support for func arg use check in SL

ZacLN:
Support for https://github.com/julia-vscode/StaticLint.jl/pull/119 (requires little ver bump for SL).
This uses  `DiagnosticTag`s which vscode represents as faded out text. These hints won't show in the problems pane. We could probably use this for other hints


davidanthoff:
StaticLint v4.1.0 is tagged, so the min bound here can be updated.

PR #679, opened 2020-05-07
update signature for SL.collect_hints

ZacLN:
Needed for https://github.com/julia-vscode/StaticLint.jl/pull/151/


davidanthoff:
StaticLint v4.1.0 is tagged, so the min bound here can be updated.

PR #683, opened 2020-05-08
update config & lint checks, remove unneeded fields

ZacLN:
- removes a few unused fields from LanguageServerInstance and Document.
- Groups the lint option fields together in LSI.
- Adds a lint option for different missing ref highlighting modes and passes this to `mark_errors`.
- Explicitly writes out all the config options we're seeking from the client. Uses new `default_options` variables in DocFormat/SL to set config (these are now all on by default). 
-Requires https://github.com/julia-vscode/StaticLint.jl/pull/160 and SL probably needs a little ver bump.


davidanthoff:
StaticLint v4.2.0 is tagged, so Project.toml here can be updated.

src/requests/workspace.jl
```diff
@@ -73,55 +73,53 @@ function process(r::JSONRPC.Request{Val{Symbol("workspace/didChangeConfiguration
...
+        ConfigurationItem(missing, "julia.lint.datadecl"),
+        ConfigurationItem(missing, "julia.lint.typeparam"),
+        ConfigurationItem(missing, "julia.lint.modname"),
+        ConfigurationItem(missing, "julia.lint.pirates"),
+        ConfigurationItem(missing, "julia.lint.useoffuncargs"),
+        ConfigurationItem(missing, "julia.lint.run"),
+        ConfigurationItem(missing, "julia.lint.missingrefs")
         ]))
+
+    new_DF_opts = DocumentFormat.FormatOptions([isnothing(opt) ? DocumentFormat.default_options[i] : opt for (i,opt) in enumerate(response[1:11])]...)
```
non-Jedi, line 103: The point of the constructors in DocumentFormat/StaticLint was so that doing these kind of checks for nothingness here is unnecessary. Is the concern that the language server receives non-Bool, non-nothing responses?

```suggestion
new_DF_opts = DocumentFormat.FormatOptions(response[1:11]...)
```

ZacLN: Sorry, I absolutely missed that - was a bit of a rush job. Could you make a PR to revert this?

non-Jedi: Sure thing.

PR #688, opened 2020-05-09
add ProgressToken constructor

### LINKED ISSUE #651: MethodError: no method matching Union{Int64, String}(::String)
I think this fails because `"initializationOptions": null,`. `LanguageServer.jl` should be able to handle this, the protocol specification says:
```
interface InitializeParams extends WorkDoneProgressParams {

...

        /**
	 * User provided initialization options.
	 */
	initializationOptions?: any;

...

}
```

Stacktrace:
```
ERROR: MethodError: no method matching Union{Int64, String}(::String)ERROR: 
Stacktrace:TypeError: in new, expected Array{Symbol,1}, got Nothing

Stacktrace: [1] LanguageServer.InitializeParams
( [1] deserialize::(Dict{String,Any}):: at Serialization/home/gkraemer/progs/julia/lsp-julia/languageserver/LanguageServer/src/protocol/initialize.jl:111.Serializer{IOStream}, 
 [2] parse_params::(DataType):: at Type/home/gkraemer/progs/julia/julia-1.4/usr/share/julia/stdlib/v1.4/Serialization/src/Serialization.jl:1315{Val{:initialize}}, 
 [2] handle_deserialize::(Dict{String,Any}):: at Serialization/home/gkraemer/progs/julia/lsp-julia/languageserver/LanguageServer/src/requests/init.jl:123.Serializer{IOStream}, 
 [3] parse::(Int32):: at Type/home/gkraemer/progs/julia/julia-1.4/usr/share/julia/stdlib/v1.4/Serialization/src/Serialization.jl:799{LanguageServer.JSONRPC.Request}, 
 [3] deserialize::(Dict{String,Any}):: at Serialization/home/gkraemer/progs/julia/lsp-julia/languageserver/LanguageServer/src/jsonrpc.jl:46.Serializer{IOStream}, 
 [4] run::(DataType):: at LanguageServerInstance/home/gkraemer/progs/julia/julia-1.4/usr/share/julia/stdlib/v1.4/Serialization/src/Serialization.jl:1309)
 at  [4] /home/gkraemer/progs/julia/lsp-julia/languageserver/LanguageServer/src/languageserverinstance.jl:236handle_deserialize
( [5] top-level scope:: at Serializationnone:1.
Serializer
Process julia-ls stderr finished
```

`lsp-mode` is sending the following:
```javascript
[Trace - 01:53:34 PM] Sending request 'initialize - (1)'.
Params: {
  "processId": null,
  "rootPath": "/home/gkraemer/.julia/dev/Base58",
  "clientInfo": {
    "name": "emacs",
    "version": "GNU Emacs 26.3 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.18.9)\n of 2019-08-30"
  },
  "rootUri": "file:///home/gkraemer/.julia/dev/Base58",
  "capabilities": {
    "workspace": {
      "workspaceEdit": {
        "documentChanges": true,
        "resourceOperations": [
          "create",
          "rename",
          "delete"
        ]
      },
      "applyEdit": true,
      "symbol": {
        "symbolKind": {
          "valueSet": [
            1,
            2,
            3,
            4,
            5,
            6,
            7,
            8,
            9,
            10,
            11,
            12,
            13,
            14,
            15,
            16,
            17,
            18,
            19,
            20,
            21,
            22,
            23,
            24,
            25,
            26
          ]
        }
      },
      "executeCommand": {
        "dynamicRegistration": false
      },
      "didChangeWatchedFiles": {
        "dynamicRegistration": true
      },
      "workspaceFolders": true,
      "configuration": true
    },
    "textDocument": {
      "declaration": {
        "linkSupport": true
      },
      "definition": {
        "linkSupport": true
      },
      "implementation": {
        "linkSupport": true
      },
      "typeDefinition": {
        "linkSupport": true
      },
      "synchronization": {
        "willSave": true,
        "didSave": true,
        "willSaveWaitUntil": true
      },
      "documentSymbol": {
        "symbolKind": {
          "valueSet": [
            1,
            2,
            3,
            4,
            5,
            6,
            7,
            8,
            9,
            10,
            11,
            12,
            13,
            14,
            15,
            16,
            17,
            18,
            19,
            20,
            21,
            22,
            23,
            24,
            25,
            26
          ]
        },
        "hierarchicalDocumentSymbolSupport": true
      },
      "formatting": {
        "dynamicRegistration": true
      },
      "rangeFormatting": {
        "dynamicRegistration": true
      },
      "rename": {
        "dynamicRegistration": true,
        "prepareSupport": true
      },
      "semanticHighlightingCapabilities": {
        "semanticHighlighting": false
      },
      "codeAction": {
        "dynamicRegistration": true,
        "isPreferredSupport": true,
        "codeActionLiteralSupport": {
          "codeActionKind": {
            "valueSet": [
              "",
              "quickfix",
              "refactor",
              "refactor.extract",
              "refactor.inline",
              "refactor.rewrite",
              "source",
              "source.organizeImports"
            ]
          }
        }
      },
      "completion": {
        "completionItem": {
          "snippetSupport": true,
          "documentationFormat": [
            "markdown"
          ]
        },
        "contextSupport": true
      },
      "signatureHelp": {
        "signatureInformation": {
          "parameterInformation": {
            "labelOffsetSupport": true
          }
        }
      },
      "documentLink": {
        "dynamicRegistration": true,
        "tooltipSupport": true
      },
      "hover": {
        "contentFormat": [
          "markdown",
          "plaintext"
        ]
      },
      "foldingRange": {
        "dynamicRegistration": true
      },
      "callHierarchy": {
        "dynamicRegistration": false
      },
      "publishDiagnostics": {
        "relatedInformation": true,
        "tagSupport": {
          "valueSet": [
            1,
            2
          ]
        },
        "versionSupport": true
      }
    },
    "window": {
      "workDoneProgress": true
    }
  },
  "initializationOptions": null,
  "workDoneToken": "1"
}
```

ZacLN:
fixes #651 ?

### DIFF
--- a/src/protocol/basic.jl
+++ b/src/protocol/basic.jl
@@ -1,4 +1,5 @@
 const ProgressToken = Union{Int, String}
+ProgressToken(x::ProgressToken) = x
 struct ProgressParams{T}
     token::ProgressToken
     value::T


jipolanco:
It seems to work. But now that I think about it, isn't this type piracy? Since `ProgressToken` is just an alias for `Union{Int, String}`...

ZacLN:
Yeah possibly, but who really owns a Union?

jipolanco:
I'd say that, since we don't own any of the types in the Union, then we don't own the Union.

If we have the right to define a constructor for the Union, then any other module has the right to do the same, which is (as I understand it) what the concept of type piracy tries to avoid.

More concretely, this implementation could break if someone else decides to define their own constructor for the same Union.

davidanthoff:
Yeah, this seems kind of risky....

ZacLN:
Probably right - I'll fix and add a check in StaticLint

gdkrmr:
why not simply do `progresstoken(x) = x` instead of `ProgressToken(x) = x`, not super pretty, but avoids Type piracy. You could also make a struct:
```
struct ProgressToken
    token::Union{String, Int}
end
```

seregaxvm:
Sorry for jumping in, but `ProgressToken` is used as a constructor only once in `src/protocol/initialize.jl` and it's not needed there. This will work just fine:
```diff
-    haskey(dict, "workDoneToken") ? ProgressToken(dict["workDoneToken"]) : missing)
+    haskey(dict, "workDoneToken") ? dict["workDoneToken"] : missing)
```

ZacLN:
Sorry, I've not had time to have a proper look at this yet. 

I'm pretty sure `ProgressToken` could be used  during the construction of any `struct` defined with the `@dict_readable` macro (these represent possible incoming messages as part of the protocol).

seregaxvm:
> I'm pretty sure `ProgressToken` could be used during the construction of any `struct` defined with the `@dict_readable` macro (these represent possible incoming messages as part of the protocol).

I've grepped project and found only this instance.
I also [compiled LanguageServer](https://github.com/julia-vscode/LanguageServer.jl/pull/695) with the patch above and it worked like a charm.

ZacLN:
The macro mentioned above creates, for each struct it is used on, functions that convert a `Dict` to that struct where keys of the Dict match the fields of the struct. The datatype specified for each field is called is called on the matching Dict value, e.g.:
```julia
macroexpand(LanguageServer, :(
@dict_readable struct T 
    d::ProgressToken 
end))
-->
struct T
    d::ProgressToken
end
function T(dict::Dict)
    T(ProgressToken(dict["d"]))
end
```

So it's not only explicit calls to `ProgressToken`s constructor that need to be adjusted, but the whole body of auto-genned code. 

Your fix above will work for now, but as soon as we start getting sent `ProgressToken`s in other contexts it will break. It's better to fix it broadly.

gdkrmr:
> The macro mentioned above creates, for each struct it is used on, functions that convert a `Dict` to that struct where keys of the Dict match the fields of the struct. The datatype specified for each field is called is called on the matching Dict value, e.g.:

Then you really should use
```
struct ProgressToken
    token::Union{String, Int}
end
```

davidanthoff:
What is the status with this? Should we first merge https://github.com/julia-vscode/LanguageServer.jl/pull/701 and then revisit this?

sdotd:
I just thought I'd mention I managed to get lsp-julia working without a hiccup by dropping in the [progtoken](https://github.com/julia-vscode/LanguageServer.jl/tree/progtoken) branch and then editing [InitializeParams](https://github.com/julia-vscode/LanguageServer.jl/issues/651#issuecomment-626127867) as suggested.  The progtoken branch did not work by itself.  I'm on mac running 1.4.2. Thanks to the guys that solved this, it was driving me crazy.

ordicker:
Could you share your init.el?

sdotd:
My lsp-mode is installed manually because I was trying different versions when I was trying to get lsp-julia to work but I don't see why an elpa installed lsp-mode wouldn't work.

[my init.el](https://github.com/sdotd/emacs.d/blob/master/init.el)

aviatesk:
@julia-vscode/core bump ? (I got a report on this today in other place) I guess it's okay if we pirate the `Union{Int,String}` constructor since LanguageServer.jl isn't supposed to interact with user code ?

non-Jedi:
I think piracy is fine in an *application*, but I'd really prefer we continue thinking of LanguageServer.jl as a library/package so that people could use parts of it to build their own code analysis tools if desired. That said, it's unlikely to happen, so if the right fix for this is onerously difficult (involves special-casing a macro? I haven't dug in to see what the best fix would be), creating a method for a `Union{Int,String}` probably won't break things.

aviatesk:
> creating a method for a Union{Int,String} probably won't break things.

na, `(::Type{Union{Int,String}})(x) = x` is essentially same as `Union{Int,String}(x) = x` and so it pirates things.

pfitzseb:
So what's the reason for not making `ProgressToken` a struct again?

aviatesk:
Guess we want to treat it as a value ?
```julia
pps = ProgressParams(...)
pps.token # want to get the actual value here ?
```
And overloading `getproperty` for every parent `struct`s who own `ProgressToken` will be verbose.

pfitzseb:
Yeah, not sure how/in what contexts `ProgressToken` is used and whether defining a bunch of `convert` methods wouldn't be enough. I can take a more thorough at some point though.

ZacLN:
If this instance of pirating is deemed unacceptable can we simply replace all fields that are typed by `::ProgressToken` with `::Union{String,Int} # ProgressToken`? I think the handwritten constructor for `InitializeParams` is the only instance where the Type is called

pfitzseb:
Sounds good to me.

aviatesk:
yeah good to me too

PR #692, opened 2020-05-10
add option to run linter on non-workspace files

ZacLN:


src/requests/textdocument.jl
```diff
@@ -300,7 +300,7 @@ function mark_errors(doc, out = Diagnostic[])
 end
 
 function publish_diagnostics(doc::Document, server)
-    if server.runlinter && server.symbol_store_ready && is_workspace_file(doc)
+    if server.runlinter && server.symbol_store_ready && (is_workspace_file(doc) || server.lint_nonwsfiles)
```
davidanthoff, line 303: Does this mean we would even publish diagnostics for files that aren't even open in the editor, but in our docs collection because we followed some `include`? That seems a bit weird to me....

I have to admit that I would still prefer that we draw a hard line and just say for files outside of the workspace we don't show diagnostics...

ZacLN: I find it frustrating to not have the linting features available in unsaved scripts and I really can't see any reason why we'd not want to provide it as an option, especially as it's defaulted to off - what's the benefit?

davidanthoff: Ah, unsafed scripts I fully agree! But we can identify those, because they don't have a `file:` URI, right? Or are they identified some other way?

But I think if someone follows the go to definition to say a file from base, then we don't really want to lint that file from base, right?

ZacLN: Yep fair. I have zero understanding of whether the URI scheme is consistent across editors - are unsaved files always prefixed with `untitled:` ?

davidanthoff: The [spec](https://microsoft.github.io/language-server-protocol/specifications/specification-current/) has one side mention of the `untitled` URI scheme, but nothing prescriptive, as far as I can tell. @non-Jedi any thoughts on this?

Maybe we just start by assuming that `untitled:` is what we handle and then see? I think in general we need to brace for more non-file URI schemes, now that VS Code is recommending git lens we will probably see quite a number of its URI schemes, and then there are also the various extensions that allow one to open a remote file system and stuff like that, which I think all will send their own URI scheme.

davidanthoff:
I think with this we don't need an option, though? Seems fair to always lint unsafed files?

src/requests/textdocument.jl
non-Jedi: "untitled" is not a valid URI scheme according to the RFC 3986 they claim to follow: https://en.wikipedia.org/wiki/List_of_URI_schemes. The mention of it is very odd.

Compounding the problem is that searching for "untitled" seems to be broken on github, so I can't even figure out if this issue has been discussed before: https://github.com/microsoft/language-server-protocol/search?q=untitled&unscoped_q=untitled

PR #696, opened 2020-05-13
interperet `nothin` -> false in client config response handling

ZacLN:
https://github.com/julia-vscode/julia-vscode/issues/1184#issuecomment-627513654
This ignores the SL/DF-Options constructors that have been recently - ths seems like an LSP specific interface so shouldn't really sit in those packages.


src/requests/workspace.jl
```diff
@@ -103,9 +103,8 @@ function request_julia_config(server::LanguageServerInstance)
         ]))
 
     server.format_options = DocumentFormat.FormatOptions(response[1:11]...)
-    server.runlinter = something(response[22], true)
+    server.runlinter = something(response[22], false)
```
non-Jedi, line 106: Well crap. Is it just this setting that vscode sends `null` for `false` on, or is it all of them? I'll go reread the spec, but this smells like a spec-violation on the part of vscode.

non-Jedi:
[Per the spec](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_configuration), a null value means that the client is unable to provide a setting which is semantically different from providing a `false` value.

> If the client can’t provide a configuration setting for a given scope then null need to be present in the returned array.

Also note that this change by itself doesn't solve https://github.com/julia-vscode/julia-vscode/issues/1184 since that issue is about "julia.lint.call" rather than "julia.lint.run".

So:

1. Is this a reported, known issue on the vs-code side?
2. Is there no way to workaround this in the vs-code plugin itself? It would be *really* nice if we didn't have to hurt spec-compliant clients because vs-code isn't complying with the spec.

non-Jedi:
One option that wouldn't break defaults for non-vscode clients would be to change all boolean config fields to be `Int`s instead with any value other than 0 being equivalent to `true`.

ZacLN:
This still needs to be verified by another vscode user I think.

ZacLN:
> Also note that this change by itself doesn't solve julia-vscode/julia-vscode#1184 since that issue is about "julia.lint.call" rather than "julia.lint.run".

Thanks - I hadn't pushed a commit

non-Jedi:
>This still needs to be verified by another vscode user I think.

Are you unable to replicate it locally?

ZacLN:
I see this behaviour on my vscode setup, not sure anyone else has confirmed it on theirs? (specifically the returning of nulls instead of false)

ZacLN:
I'd rather we changed it to "on"/"off" to be a bit more transparent - is that ok with you?

non-Jedi:
That's fine with me.

davidanthoff:
> I'd rather we changed it to "on"/"off" to be a bit more transparent

Should we maybe first check with the MS guys whether there is some other way around this? If our current understanding of the situation is correct, I agree that "on/off" is a better option, but this is frustrating that one can't just use a `bool` and be done with it... Maybe we could even convince them to just change the implementation and fix this?

Changing all of these settings to "on/off" is also annoying because every user who adjusted those settings will have to update...

marius311:
Just want to confirm that for me this PR fixes the inability to turn linting off that I mentioned on Slack and is mentioned in https://github.com/julia-vscode/julia-vscode/issues/1184.

Would be great to get a fix in if possible, since this is a real "paper cut" bug I think alot of new users would probably hit.

ZacLN:
OK, this definitely looks like a [bug](https://github.com/microsoft/vscode-languageserver-node/issues/623) on the vscode side of things. I'm going to close this and put forward a pair of PRs for LanguageServer and julia-vscode that would move these configurations from bools to "on"/"off". 

This feels like something that should be addressed rather quickly and I don't have a sense of how quickly changes would make their way through upstream. It's then an easy change to revert back when the bug is addressed. Thoughts?

non-Jedi:
@ZacLN why not just use this branch to do a release for vscode (and fix the obviously super obnoxious ux of not being able to turn off configurration) and then wait and see if the upstream bug is quickly fixed before you next want to do a real LanguageServer.jl release? That way we potentially avoid people having to redo their config.

ZacLN:
I'm amenable to that, @davidanthoff ?

davidanthoff:
Yes, I would much prefer that we stick with `bool` and `true`/`false`, and my understanding is that this PR here works around the bug in VS Code, right? So I would go with that strategy for now.

Normally up stream is super fast in addressing stuff. I'm sure if we were to open a PR ourselves that fixes the problem, it could be shipped very quickly. Also, because this is a bug in the `vscode-languageclient` npm package, we might not even have to wait for a VS Code release, I think they can release a bug fix release of that package independently from VS Code releases.

Also very much agree that we should ship this ASAP!

ZacLN:
I think @non-Jedi's proposal was that we use this branch in the snapshot for a julia-vscode release rather than merging this

davidanthoff:
I don’t think we can ship a branch, there is a fair bit of tooling in place on the extension side that assumes we ship `master`, and with our regular releases it would be painful to work around that. 

Can we put the content of this branch behind an `if` clause that checks `InitializeParams. clientInfo.name` so that we only use this implementation if the client is VS Code? Presumably VS Code sends that field?

But in any case, we should ship something ASAP, right now users on the released version can’t change any of these settings and we also have crash reports because somewhere there is a call to a method with the wrong number of args in the currently shipping version.

@ZacLN let me know if you want me to do something about this PR if you are busy!

ZacLN:
OK, I've amended as above

PR #697, opened 2020-05-14
add "julia/getCurrentModule" handler

ZacLN:
Handler for "julia/getCurrentModule" requests from the client. Takes a `TextDocumentPositionParams` and returns a string of the current module.


src/requests/features.jl
```diff
@@ -305,3 +305,22 @@ function _binding_kind(b ,server)
         return 13
     end
 end
+
+JSONRPC.parse_params(::Type{Val{Symbol("julia/getCurrentModule")}}, params) = TextDocumentPositionParams(params)
+function process(r::JSONRPC.Request{Val{Symbol("julia/getCurrentModule")},TextDocumentPositionParams}, server)
```
davidanthoff, line 310: There isn’t really a sense of `current` in this API, right? Maybe `getModuleAt` or `getSurroundingModule` or something like that?

ZacLN: Makes sense - also are we happy to use "julia/..." for these arbitrary requests?

PR #699, opened 2020-05-16
Fix test_path tests

pxl-th:
- Add tests for paths to `runtests.jl`.
- Fix tests for paths.

PR #702, opened 2020-05-17
Add escaping when comparing/hashing URI2

pxl-th:
- Add function for escaping URI
- Add tests for escaping/comparing/hashing URI

This is a continuation of discussion in #699 PR, that allows comparing URIs that were created using different encoding schemes.
Basically what it does is it escapes URI to bring it to the same encoding scheme as in [vscode-uri](https://github.com/Microsoft/vscode-uri).
This allows for following comparisons:
```julia
hash(LanguageServer.URI2("file:///D:/FOO/bar")) == hash(LanguageServer.URI2("file:///d%3A/FOO/bar"))
```

It escapes only file URI, since I'm not sure if escaping is needed in other cases.

**Note:** It still needs to be checked if _"VS Code internally also treats two strings that stand for the same URI but use different encoding patterns as equal or not"_, so this may be considered as WIP.

### DIFF
--- a/src/uri2.jl
+++ b/src/uri2.jl
@@ -11,23 +11,23 @@ import Base.==
 function ==(a::URI2, b::URI2)
     @static if Sys.iswindows()
         if startswith(a._uri, "file://")
-            return lowercase(a._uri) == lowercase(b._uri)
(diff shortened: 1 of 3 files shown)


davidanthoff:
Thanks!

Issue #703, opened 2020-05-17
Possible document duplicates


pxl-th:
`LanguageServerInstance` contains documents in `_documents` dict, that have different URI keys, but point to the same document, here's an example obtained by `getdocuments_key(server)` command:
```
[ Info: Server documents: LanguageServer.URI2[
    LanguageServer.URI2(\"file:///SigmaPoints.jl\"),
    LanguageServer.URI2(\"file:///c%3A/Users/tonys/projects/julia/KalmanFilters/src/KalmanFilters.jl\"),
    LanguageServer.URI2(\"file:///Helpers.jl\"),
    LanguageServer.URI2(\"file:///UnscentedKalmanFilter.jl\"),
    LanguageServer.URI2(\"file:///c%3A/Users/tonys/projects/julia/KalmanFilters/src/Helpers.jl\"),
    LanguageServer.URI2(\"file:///c%3A/Users/tonys/projects/julia/KalmanFilters/src/UnscentedKalmanFilter.jl\"),
    LanguageServer.URI2(\"file:///c%3A/Users/tonys/projects/julia/KalmanFilters/src/SigmaPoints.jl\")]",""
]]
```
Both `LanguageServer.URI2(\"file:///UnscentedKalmanFilter.jl\")` and `LanguageServer.URI2(\"file:///c%3A/Users/tonys/projects/julia/KalmanFilters/src/UnscentedKalmanFilter.jl\")` are pointing to the same file, but having different URI.

I've found out that URIs with "relative" path are added in [`set_file`](https://github.com/julia-vscode/LanguageServer.jl/blob/master/src/staticlint.jl#L25) method.

Is this an intended behavior? Becase the error message in `set_file` method itself says, that
`"StaticLint should not try to set documents that are already tracked."`.

**Steps to reproduce:**
1. Add `@info "Server documents: $(getdocuments_key(server))"` to the main loop in `Base.run` method for `LanguageServer`, for example.
2. Open some julia file and do nothing else, just watch the logs.

Not sure if this occurs in vscode, but this was consistent for me when opening different files from different projects in Vim.

**Setup:**
Windows 10, Vim, vim-lsp

Btw, I found this ?issue? while trying to figure out why another issue occured to me.

davidanthoff:
This is definitely a bug, we should just never use relative path URIs. I take it that those relative URIs are generated somewhere in LS or StaticLint.jl? Then we should fix that.

ZacLN:
[...]

PR #704, opened 2020-05-17
Add check for relative paths

davidanthoff:
Helps diagnose #703.

The general idea is that we want to figure out where relative paths come from, and this might help us to narrow that scenario down. My understanding is that we will never receive a URI from a client with a relative path, but that those relative paths originate somewhere in our own code, maybe in the include following logic of StaticLint? Of course, I would just generally remove any loading of non-workspace files from disc :)

CC @pxl-th and @non-Jedi.

### DIFF
--- a/src/exception_types.jl
+++ b/src/exception_types.jl
@@ -45,3 +45,11 @@ end
 function Base.showerror(io::IO, ex::LSPositionToOffsetException)
     print(io, ex.msg)
 end
+
+struct LSRelativePath <:Exception
+    msg::AbstractString
+end
+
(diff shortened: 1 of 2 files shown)


ZacLN:
It's almost certainly coming from within StaticLint but this isn't related to whether a file is in the workspace or not. I don't think this PR would give us the info we need (it'll notify of the issue but not how it's come about), I'll make one for SL

davidanthoff:
We should probably still merge this, right? It can't harm and essentially would make sure we would notice any problem in this area.

PR #720, opened 2020-05-27
don't suggest gensymed names for completions

### LINKED ISSUE #719: Autocompletions should not contain invalid names
image

ZacLN:
This should prevent gensymed variables being suggested for completions. This reminds me that completions don't works with the `var"someid"` syntax but that's not an immediate priority

closes #719


ZacLN:
While we're on this - what other completions do we want to omit that are currently provided but may be surprising? For example, I think we will currently be providing completions for some imported variables not defined within a module, these should probably be omitted as well, right?

aviatesk:
> currently be providing completions for some imported variables not defined within a module, these should probably be omitted as well, right?

I think imported variables should be completed, no ? REPL doesn't provide completions on `using`ed variable/methods, but that has been my minor gripe to say honestly.

PR #724, opened 2020-05-27
Add isnothing for Julia < 1.1

davidanthoff:
Leading to bugs in the wild right now, we should ship quickly.


non-Jedi:
For reference, if `isnothing` is ever in a performance critical section of code, `=== nothing` should be preferred anyway: https://github.com/JuliaLang/julia/issues/27681

`isnothing` is often more convenient/terse, so I'm definitely not advocating avoiding it, but I found this surprising when I learned, so I thought it worth mentioning.

PR #726, opened 2020-05-29
fix off by one in sig helper

### LINKED ISSUE #725: Method hover help off by one argument
image

Should show both the two- and three-arg methods with only two arguments specified (or maybe only the two-arg version unless there's a trailing comma).

ZacLN:
closes #725


davidanthoff:
I think the Mac github runners are generally down, so I'm merging without them passing.

PR #727, opened 2020-05-29
pre-empt incoming staticlint change

ZacLN:
Behaviour is equivalent


davidanthoff:
Merging, because the remaining CI runs are Mac, which seems generally down.

Issue #730, opened 2020-05-30
LanguageClient-neovim can't talk to LanguageServer.jl


ExpandingMan:
I'm not sure why, all I can see is `[LC] Sending on a disconnected channel` and soon times out.

I'm running the following
```vim
let g:LanguageClient_serverCommands = {
\   'julia': ['julia', '--startup-file=no', '--history-file=no', '-e', '
\       using LanguageServer;
\       using Pkg;
\       import StaticLint;
\       import SymbolServer;
\       env_path = dirname(Pkg.Types.Context().env.project_file);
\       server = LanguageServer.LanguageServerInstance(stdin, stdout, env_path);
\       server.runlinter = true;
\       run(server);
\   ']
\ }
```
I've checked that this script runs fine in Julia directly, but I really have no idea how to see what's going on through neovim or diagnose what's happening.

Any suggestions?  I'd be happy to provide more information, but I've gotten rather stuck here as I really can't figure out where to look.

davidanthoff:
I don't know whether that is it, but in the VS Code extension we redirect stdout like [this](https://github.com/julia-vscode/julia-vscode/blob/master/scripts/languageserver/main.jl#L24).

ExpandingMan:
Hm, well I've tried that trick by changing the script to
```
let g:LanguageClient_serverCommands = {
\   'julia': ['julia', '--startup-file=no', '--history-file=no', '-e', '
\       using LanguageServer;
\       using Pkg;
\       import StaticLint;
\       import SymbolServer;
\       cnxn = stdout
\       redirect_stdout()
\       env_path = dirname(Pkg.Types.Context().env.project_file);
\       server = LanguageServer.LanguageServerInstance(stdin, cnxn, env_path);
\       server.runlinter = true;
\       run(server);
\   ']
\ }
```
I've also gotten some dumps of logs from LanguageClient-neovim, although I still don't know what to make of them, the error lines are
```
20:51:33 ERROR reader-Some("julia") src/rpcclient.rs:35 Thread reader-Some("julia") exited with error: ErrorMessage { msg: "Unable to read from language server" }
20:52:03 ERROR writer-Some("julia") src/rpcclient.rs:46 Thread writer-Some("julia") exited with error: Os { code: 32, kind: BrokenPipe, message: "Broken pipe" }
20:52:18 ERROR unnamed src/rpchandler.rs:13 Error handling message: sending on a disconnected channel
[...]

PR #731, opened 2020-05-30
possible workaround for clients that incorrectly pass null values

ExpandingMan:
This is a possible workaround for #730 .

Please only feel compelled to merge this if you really think it's appropriate, and not just because you hope it'll get things working, because it won't.  Even with this issue resolved all sorts of (seemingly unrelated) things are going wrong when trying to use this with neovim, so this absolutely does not get it working, and this has been quite an exercise in frustration.

### DIFF
--- a/src/protocol/protocol.jl
+++ b/src/protocol/protocol.jl
@@ -43,7 +43,11 @@ macro dict_readable(arg)
             if fieldtype isa Expr && fieldtype.head == :curly && fieldtype.args[2] != :Any
                 f = :($(fieldtype.args[2]).(dict[$fieldname]))
             elseif fieldtype != :Any
-                f = :($(fieldtype)(dict[$fieldname]))
+                f = if field_allows_missing(field)
+                    :($fieldtype(dict[$fieldname] isa Nothing ? missing : dict[$fieldname]))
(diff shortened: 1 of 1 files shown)


ZacLN:
Hiya, I'm keen to help get nvim working but I'd prefer a much narrower workaround than this. 

One thing I'm not clear on is whether this inconsistency is specific to this part of the LSP specification or whether it sends nulls in place of missing fields across the board (I had a look at the client code but couldn't understand it).

davidanthoff:
Maybe this will be fixed in the client, they just merged something https://github.com/gluon-lang/lsp-types/pull/160.

ExpandingMan:
Yeah, you just beat me to it, I was about to mention that we just got [this response](https://github.com/autozimu/LanguageClient-neovim/issues/940#issuecomment-636351877).

I think my vote would be not to merge this.  As I indicated, the interaction with neovim seems very broken, I doubt it'll be fixed anytime soon unless somebody dedicates their entire weekend to it and makes PR's to both repositories, unfortunately.  In light of that, it doesn't seem worth it to start going down the whole of compensating for clients that don't follow the standard.

PR #738, opened 2020-06-02
Refactor -> anonymous function

non-Jedi:
Using the `function(x)` syntax makes this callback clearer and doesn't
confuse indentation so much.

### DIFF
--- a/src/languageserverinstance.jl
+++ b/src/languageserverinstance.jl
@@ -170,7 +170,7 @@ function trigger_symbolstore_reload(server::LanguageServerInstance)
         ssi_ret, payload = SymbolServer.getstore(
             server.symbol_server,
             server.env_path,
-            function(i)
+            function (i)
                 if server.clientcapability_window_workdoneprogress && server.current_symserver_progress_token !== nothing
                     JSONRPC.send(
                         server.jr_endpoint,
--- a/src/languageserverinstance.jl
+++ b/src/languageserverinstance.jl
@@ -170,15 +170,17 @@ function trigger_symbolstore_reload(server::LanguageServerInstance)
         ssi_ret, payload = SymbolServer.getstore(
             server.symbol_server,
             server.env_path,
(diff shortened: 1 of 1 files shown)


davidanthoff:
Hm, we've never used that syntax, and I think it is not even mentioned in the Julia manual?

The indentation problem (if the current one is the result of DocumentFormat.jl) seems like a bug, right?

non-Jedi:
> Hm, we've never used that syntax, and I think it is not even mentioned in the Julia manual?

[It's introduced in the manual at the same place as `->` anonymous functions.](https://docs.julialang.org/en/v1/manual/functions/#man-anonymous-functions-1)

> The indentation problem (if the current one is the result of DocumentFormat.jl) seems like a bug, right?

Sure. I think this way is much cleaner, but we can just fix the indentation instead. Someone else will have to do that though because I (and my editor) honestly have no idea what the "correct" indentation would be.

davidanthoff:
Ah, I didn't find that in the docs because I searched for the version without a space between `function` and `(` :) Ok, then I'm on board, but should we then also add a space between the `function` keyword and the arg list, like in the docs?

non-Jedi:
What's y'all's procedure for doing merges? Once someone has approved a PR and CI has passed can it be merged any time? Or in general should I let you and Zac do merges so I don't accidentally merge something when y'all are planning for a release as mentioned in https://github.com/julia-vscode/SymbolServer.jl/pull/160#issuecomment-636985981?

davidanthoff:
So right now we are playing with a model where only I merge things, because I coordinate the VS Code releases, see https://github.com/julia-vscode/julia-vscode/wiki/Dev-process.

But that is not set in stone, we're still experimenting right now.

But I think for now, just assume that if you author a PR and it is not in draft mode, it might be merged at any time (if someone approves it and I merge it), i.e. as an author, once it is not a draft anymore, it is out of your control ;) For other PRs, best if you review and approve, and then I'll merge those so that it fits with the VS Code release.

PR #740, opened 2020-06-05
improve keyword snippets

aviatesk:
- rm dead keywords
- use LSP's snippet-tab instead of hardcoded 4 spaces, which will automatically respect current editor spaces/tabs seeting

### DIFF
--- a/src/requests/completions.jl
+++ b/src/requests/completions.jl
@@ -63,103 +63,87 @@ function latex_completions(doc, offset, partial, CIs)
     end
 end
 
-function kw_completion(doc, spartial, ppt, pt, t, CIs, offset)
-    length(spartial) == 0 && return
-    fc = first(spartial)
-    if startswith("abstract", spartial)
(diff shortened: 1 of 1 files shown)


ZacLN:
Could we try a terser approach such as: 
```julia
const snippet_completions = Dict{String,String}(
    "abstract" => "abstract type \$0 end",
    "baremodule" => "baremodule \$0\nend",
    "begin" => "begin\n    \$0\nend", 
    "break" => "break",
    ...)
function kw_completion(doc, spartial, ppt, pt, t, CIs, offset)
    length(spartial) == 0 && return
    fc = first(spartial)
    for (kw,comp) in snippet_completions
        if startswith(kw, spartial)
            push!(CIs, CompletionItem(kw, 14, kw, TextEdit(Range(doc, offset:offset), comp[length(spartial) + 1:end])))
        end
    end
end
```

This keeps all the snippets groups together and a single bit of code for adding them (avoiding use of a macro).

Does that sound OK?

davidanthoff:
We can close this because https://github.com/julia-vscode/LanguageServer.jl/pull/761 is taking its place, right? If not, please reopen!

PR #743, opened 2020-06-08
method jump in hover

aviatesk:
works on top of https://github.com/julia-vscode/SymbolServer.jl/pull/166
(note vscode doesn't seem to handle `file:line` case, so we can only jump to a file)

context: https://github.com/julia-vscode/SymbolServer.jl/pull/166#issuecomment-640667882


aviatesk:
will work like this:
image

pfitzseb:
Let's get rid of the `[x]` numbers and put the function name into backticks too?

src/requests/hover.jl
```diff
@@ -66,12 +66,33 @@ function get_hover(b::StaticLint.Binding, documentation::String, server)
...
+# print(io, x::SymStore) methods are defined in SymbolServer
 function get_hover(b::SymbolServer.SymStore, documentation::String, server)
     if !isempty(b.doc)
         documentation = string(documentation, b.doc, "\n")
     end
     documentation = string(documentation, "```julia\n", b, "\n```")
-    # print(io, x::SymStore) methods are defined in SymbolServer
+end
+
+const JULIA_DIR = normpath(joinpath(Sys.BINDIR, Base.DATAROOTDIR, "julia"))
```
davidanthoff, line 77: We'll have to think hard about this one here because I think it might make the LS code non-relocatable, which will cause trouble when we want to statically compile it.

Maybe an easier way would be that we fix these paths in the symbol server process? That will always run in a Julia instance that "makes sense". But I'm not sure that works... @ZacLN?

aviatesk:
> Let's get rid of the [x] numbers and put the function name into backticks too?

how about 
image

?

aviatesk:
Now it doesn't need  julia-vscode/SymbolServer.jl#166, I ended up using an equivalent logic here as well.

src/requests/hover.jl
pfitzseb: We can just set this at package load time, not at precompilation time (which we don't do at the moment anyways).

davidanthoff: I think that will still be wrong: say we statically compile LS with Julia 1.4.2, but the user has VS Code configured to use Julia 1.0.5. So in this case the LS process will run in Julia 1.4.2, but we want paths to the local install of Julia 1.0.5 of the user, which we won't get in the LS process. But we _would_ get those correct paths in the symbol server process because that would run under Julia 1.0.5.

ZacLN: Is there an issue with the paths from SymbolServer other than that we don't want to display them here?

davidanthoff: I think (but not sure) the goal here is simply to return absolute paths for the files from base? I think we should easily be able to make that transformation in the symbol server.

aviatesk: yes, I basically tried to mimic how `Base` displays methods:
```julia
julia> methods(sin)
# 14 methods for generic function "sin":
[1] sin(x::BigFloat) in Base.MPFR at mpfr.jl:727
[2] sin(a::Complex{Float16}) in Base.Math at math.jl:1146
[3] sin(a::Float16) in Base.Math at math.jl:1145
[4] sin(::Missing) in Base.Math at math.jl:1198
...
```

> I think we should easily be able to make that transformation in the symbol server.

We need absolute file path here since we want to create links in this package, so I guess the easiest workaround is just to drop this feature ? Or defining `JULIA_DIR` in SymbolServer.jl and use that here, maybe ?

davidanthoff: Yes, I think the latter: `JULIA_DIR` in the symbol server process will point to the correct location, even once we move the LS over to a statically compiled version.

aviatesk:
needs https://github.com/julia-vscode/SymbolServer.jl/pull/168 to work

aviatesk:
final look
image

src/requests/hover.jl
davidanthoff: This line can be delete now, right?

src/requests/hover.jl
```diff
@@ -66,12 +66,45 @@ function get_hover(b::StaticLint.Binding, documentation::String, server)
...
+            if sig[1] ≠ Symbol("#unused#")
+                print(io, sig[1])
+            end
+            print(io, "::", sig[2])
+            i ≠ nsig && print(io, ", ")
+        end
+        print(io, ")")
+        sig = String(take!(io))
+        mod = m.mod
+        text = string(replace(m.file, SymbolServer.JULIA_DIR => ""), ':', m.line)
```
davidanthoff, line 101: I think this won't work with the statically compiled LS server: it is accessing the JULIA_DIR var in the LS process, right?

The key thing is that we _only_ use the dir for base that we get from querying `Sys.BINDIR` and `Base.DATAROOTDIR` in the symbol server process.

aviatesk: still not sure how I can solve what you're concerning.
After all, (statically compiled) LS process can't know which version of Julia an user wants to interacts with, no ?

aviatesk: couple of questions:
- will SymbolServer.jl be statically compiled ?
- if SymbolServer compiled, and it will replace its `JULIA_DIR` at runtime (i.e. given based on user-configured julia), then `SymbolServer.JULIA_DIR` here won't point at that ? Will static compiler will resolve constants in other modules at compile time ?

If questions above are true, I agree that the easiest way to implement this would be replace this kind of stuff in SymbolServer, and keep that in additional `MethodStore`'s store.
But imho the field doesn't deserve its complexity (non-Base functions don't need that after all), so I'm totally okay to remove these replacements from the beginning. It's just that it will just look verbose.

davidanthoff: > will SymbolServer.jl be statically compiled ?

The part in SymbolServer.jl that is running inside the LS process will be statically compiled, but the symbol server process won't, we will just continue to start a Julia instance for that.

Re additional stuff in the `MethodStore`, I'm not sure, probably @ZacLN has a better sense of that.

aviatesk: for now, I just remove this replacement.

davidanthoff: Sounds good!

davidanthoff:
Can you merge `master` into this PR? You can just click "Update branch" to do that.

And then I think @ZacLN needs to review this.

aviatesk:
> Can you merge master into this PR? You can just click "Update branch" to do that.

Done ;) Requested review from Zac.

davidanthoff:
I'll continue merging things like this until we are done with all the showstopper bugs, at which point we should probably put some sort of merge freeze in place until v0.16 is out.

PR #744, opened 2020-06-09
julia/getDocAt handler

aviatesk:
supersedes #377 

currently just borrows logics of `get_hover`, and many TODOs left:
~- [ ] handle links (e.g. `@ref`)~
~- [ ] horizontal line separators~
~- [ ] documentations on variables, structs~
~- [ ] more accurate documentation retrieve (e.g. `Base.sin`)~

Let's address those TODOs later, hacking in LS will take some time.


src/requests/features.jl
```diff
@@ -317,3 +317,38 @@ function get_module_of(s::StaticLint.Scope, ms = [])
         return isempty(ms) ? "Main" : join(ms, ".")
     end
 end
+
+using Base.Docs, Markdown
+using Markdown: MD, HorizontalRule
```
davidanthoff, line 322: I think we have a convention in the codebase to have all `using` statements in the root `LanguageServer.jl` file.

Project.toml
```diff
@@ -3,35 +3,36 @@ uuid = "2b0e0bc5-e4fd-59b4-8912-456d1b03d8d7"
...
-JSONRPC = "1"
+julia = "1"
+
+[extras]
+LibGit2 = "76f85450-5226-5b5a-8eaa-529ad045b433"
+Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
+SHA = "ea8e919c-243c-51af-8825-aaa63cd721ce"
+Serialization = "9e88b42a-f829-5b0c-bbe9-9e923198166b"
+Sockets = "6462fe0b-24de-5631-8697-dd941f90decc"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
```
davidanthoff, line 35: Could this be done with less reorganization? I think this PR only adds one package, right?

I also think (but not 100% positive) that PkgDev.jl (which I use for tagging versions) will change things back to whatever order it likes, so we shouldn't try to "fight" the order here, but just leave as is and ideally just add individual lines.

aviatesk: sounds fair, `Pkg.add("Markdown")` automatically reorders things, so maybe we should have think about making PkgDev.jl to follow that though.

src/protocol/messagedefs.jl
```diff
@@ -24,6 +24,7 @@ const workspace_applyEdit_request_type = JSONRPC.RequestType("workspace/applyEdi
 const workspace_configuration_request_type = JSONRPC.RequestType("workspace/configuration", ConfigurationParams, Vector{Any})
 
 const julia_getModuleAt_request_type = JSONRPC.RequestType("julia/getModuleAt", TextDocumentPositionParams, String)
+const julia_getDocAt_request_type = JSONRPC.RequestType("julia/getDocAt", TextDocumentPositionParams, String)
```
davidanthoff, line 27: Should we use the struct here that has the version information included that @pfitzseb just added to the code base? We're potentially going to run into the same issue here, right?

davidanthoff:
@aviatesk needs a merge conflict resolution.

src/protocol/messagedefs.jl
pfitzseb: Probably a good idea, yes.

aviatesk: What should we return as `JSONRPCError`'s `data` field ? `getModuleAt` and `getBlockAt` return `nothing`.

src/requests/features.jl
```diff
@@ -376,3 +376,44 @@ function get_module_of(s::StaticLint.Scope, ms = [])
...
+function julia_getDocAt_request(params::VersionedTextDocumentPositionParams, server::LanguageServerInstance, conn)
+    uri = URI2(params.textDocument.uri)
+    hasdocument(server, uri) || return nodocuemnt_error(uri)
+
+    doc = getdocument(server, uri)
+    if doc._version !== params.version
+        return mismatched_version_error(uri, doc, params, "getDocAt")
+    end
+
+    x = get_expr1(getcst(doc), get_offset(doc, params.position))
```
pfitzseb, line 391: Shouldn't this be `get_expr`?

src/protocol/messagedefs.jl
davidanthoff: I think `nothing` is fine, not sure what one is supposed to do with that field ;)

aviatesk: hehe, I get it :)

src/requests/features.jl
aviatesk: hm, why do you think so ?
I adapted code from [this line](https://github.com/julia-vscode/LanguageServer.jl/blob/0198aeec2a1310385e431be0e1901f36f3756855/src/requests/hover.jl#L3), and not so sure about the difference between `get_expr` and `get_expr1`...
Maybe @ZacLN could you explain this to me ?

PR #755, opened 2020-06-14
alter getCurrentBlockRange request

### LINKED ISSUE #716: 'julia/getCurrentBlockRange' selects everything up to the current line when on empty line
Should just return an empty range in this case, imho.

ZacLN:
fixes #716 ?

Changes the behaviour to ignore trailing white space when considering which block to retrieve. By default returns an empty selection at (0,0):(0,0). @pfitzseb , is this what was wanted?

### DIFF
--- a/src/requests/misc.jl
+++ b/src/requests/misc.jl
@@ -5,47 +5,49 @@ end
 function setTraceNotification_notification(params, server::LanguageServerInstance, conn)
 end
 
-function julia_getCurrentBlockRange_request(params::TextDocumentPositionParams, server::LanguageServerInstance, conn)
-    tdpp = params
+function julia_getCurrentBlockRange_request(tdpp::TextDocumentPositionParams, server::LanguageServerInstance, conn)
     doc = getdocument(server, URI2(tdpp.textDocument.uri))
     offset = get_offset(doc, tdpp.position)
     x = getcst(doc)
     loc = 0
-    p1, p2, p3 = 0, x.span, x.fullspan
+
     if typof(x) === CSTParser.FileH
-        (offset > x.fullspan || x.args === nothing) && return Position(get_position_at(doc, p1)...), Position(get_position_at(doc, p2)...), Position(get_position_at(doc, p3)...)
         for a in x.args
-            if loc <= offset < loc + a.fullspan
-                if typof(a) === CSTParser.ModuleH
-                    if loc + a.args[1].fullspan + a.args[2].fullspan < offset < loc + a.args[1].fullspan + a.args[2].fullspan + a.args[3].fullspan
-                        loc0 = loc +  a.args[1].fullspan + a.args[2].fullspan
+            if loc <= offset <= loc + a.span
+                if CSTParser.defines_module(a) # Within module at the top-level, lets see if we can select on of the block arguments
+                    if loc <= offset <= loc + a.args[1].span # Within `module` keyword, so return entire expression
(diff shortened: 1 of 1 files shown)


src/requests/misc.jl
```diff
@@ -5,47 +5,49 @@ end
...
     if typof(x) === CSTParser.FileH
-        (offset > x.fullspan || x.args === nothing) && return Position(get_position_at(doc, p1)...), Position(get_position_at(doc, p2)...), Position(get_position_at(doc, p3)...)
         for a in x.args
-            if loc <= offset < loc + a.fullspan
-                if typof(a) === CSTParser.ModuleH
-                    if loc + a.args[1].fullspan + a.args[2].fullspan < offset < loc + a.args[1].fullspan + a.args[2].fullspan + a.args[3].fullspan
-                        loc0 = loc +  a.args[1].fullspan + a.args[2].fullspan
+            if loc <= offset <= loc + a.span
+                if CSTParser.defines_module(a) # Within module at the top-level, lets see if we can select on of the block arguments
+                    if loc <= offset <= loc + a.args[1].span # Within `module` keyword, so return entire expression
```
aviatesk, line 18: This would be really helpful, while we still want https://github.com/julia-vscode/julia-vscode/issues/1330 to be solved for great module experience.

src/requests/misc.jl
```diff
@@ -5,47 +5,49 @@ end
...
-                                p1, p2, p3 = loc, loc + b.span, loc + b.fullspan
-                                break
+                            if loc <= offset <= loc + b.span
+                                return Position(get_position_at(doc, loc)...), Position(get_position_at(doc, loc + b.span)...), Position(get_position_at(doc, loc + b.fullspan)...)
                             end
                             loc += b.fullspan
                         end
-                    else
-                        p1, p2, p3 = loc, loc + a.span, loc + a.fullspan
+                    elseif loc + a.args[1].fullspan + a.args[2].fullspan + a.args[3].fullspan < offset < loc + a.args[1].fullspan + a.args[2].fullspan + a.args[3].fullspan + a.args[4].span # Within `end` of the module, so return entire expression
```
aviatesk, line 33: so this branch tried to return the entire module code at when we query request in `end` keyword, right ?

I tried the code below manually and found this branch seeming never got hit ?
```julia
module A
a rand(Bool)
end|
```
where `|` represents the request position.

aviatesk:
Will fix https://github.com/julia-vscode/julia-vscode/issues/1281 as well

src/requests/misc.jl
```diff
@@ -5,47 +5,49 @@ end
...
                 else
-                    p1, p2, p3 = loc, loc + a.span, loc + a.fullspan
+                    return Position(get_position_at(doc, loc)...), Position(get_position_at(doc, loc + a.span)...), Position(get_position_at(doc, loc + a.fullspan)...)
                 end
             end
             loc += a.fullspan
         end
     end
-    return Position(get_position_at(doc, p1)...), Position(get_position_at(doc, p2)...), Position(get_position_at(doc, p3)...)
+    return Position(0, 0), Position(0, 0), tdpp.position
```
pfitzseb, line 50: Could we try to get the next block here as well instead of returning the current cursor position?

davidanthoff:
We probably need to merge some version of this before a release, right?

src/requests/misc.jl
pfitzseb: That case works fine for me.

pfitzseb:
Yes. Let's merge this as-is -- the inline comments aren't a big deal and can imho be addressed in a follow-up PR since this fixes a pretty annoying bug.

PR #756, opened 2020-06-14
fix FormatOptions inconsistency

ZacLN:
The wrong configuration name was being requested, `kwspacing -> kwarg`.

Mentioned in https://github.com/julia-vscode/LanguageServer.jl/pull/753


davidanthoff:
Haha, I just opened PRs against DocumentFormat and julia-vscode that uses `kwspacing` for this.

`kwspacing` seems a more descriptive name for the config setting, right, so should we go with that?

ZacLN:
Happy either way (I was clearly in two minds at the time!). Using kwspacing will require a PR against `package.json` in julia-vscode

davidanthoff:
You decide, this is more your part of the codebase :) The PRs that are needed for `kwspacing` are also open, you can just either merge those two, or this one here.

PR #759, opened 2020-06-16
RFC: Use JuliaFormatter instead of DocumentFormat

non-Jedi:
Closes https://github.com/julia-vscode/DocumentFormat.jl/issues/70

JuliaFormatter was originally forked from DocumentFormat when @domluna
wanted to add canonicalising functionality. This resulted in a
large-scale refactoring that could not be immediately merged to
DocumentFormat (https://github.com/julia-vscode/DocumentFormat.jl/pull/27).
Since then, extensive work has gone into JuliaFormatter. I believe it is
significantly more feature-rich than DocumentFormat at this time.

I believe there are significant advantages both to the Julia community
at large and to LanguageServer.jl to combine efforts on a single code
formatter.

Possible objections (I'm updating this list as discussions continue):

- [ ] Missing features from DocumentFormat not present in JuliaFormatter.
  - I am not aware of any but this should be more thoroughly reviewed.
- [ ] JuliaFormatter is more aggressive than desired (e.g. by inserting line breaks).
  - My understanding is that this could be mitigated using JuliaFormatter's style system. Doing so would require more work before merging this.
  - WIP DocumentFormat.jl style at https://github.com/domluna/JuliaFormatter.jl/pull/256
- [ ] Lack of julia-vscode governance.
  - Taking a dependency on something not controlled by julia-vscode members may not be desired for reasons of making sure necessary features and fixes for vscode releases are merged in a timely manner.
- [ ] JuliaFormatter has an additional dependency on DataStructures. My understanding is that this is undesirable for the way the vscode extension is packaged.
- [ ] [Stability concerns (crashing) with JuliaFormatter vs DocumentFormat](https://github.com/julia-vscode/julia-vscode/issues/1309#issuecomment-640313314)


domluna:
I'd like to point out a style that's very similar to what DocumentFormat
currently does is possible. I just haven't gotten around to implementing it
as of yet.

non-Jedi:
Thanks for chiming in @domluna, if you could point out which methods would need to be defined for a custom style in order to prevent inserting line breaks (since that was a point of contention in https://github.com/julia-vscode/DocumentFormat.jl/issues/70, that would probably be helpful for this discussion. Or if it's more or less all of them, that's also valuable information.

pfitzseb:
I'm in favour of this.

David has [some concerns](https://github.com/julia-vscode/julia-vscode/issues/1309#issuecomment-640313314) about this though. AFAICT JuliaFormatter is more widely used and more powerful than DocumentFormat, so imho it's a good idea regardless.

non-Jedi:
I've added that concern to the OP.

My opinion is that even if there are regressions due to this change, JuliaFormatter provides enough additional features that making the change is worthwhile. Just in the past couple hours of using this branch, I've found myself actually making use of the formatting feature--which I had previously largely ignored--because it's so much better.

I believe getting DocumentFormat to have the same level of features as JuliaFormatter would introduce more bugs than currently exist in JuliaFormatter.

src/formatoptions.jl
```diff
@@ -0,0 +1,16 @@
...
+FormatOptions() = FormatOptions(default_format_options...)
+
+FormatOptions(options::Vararg{Any,length(default_format_options)}) =
+    FormatOptions(something.(options, default_format_options)...)
+
+JuliaFormatter.getstyle(x::FormatOptions) = x
+
+# All functions that don't have a dispatch defined for FormatOptions
+# fallback to the definition for JuliaFormatter.DefaultStyle. This is
+# how we customize the behavior of the formatter.
```
aviatesk, line 16: I would like to first make sure whether we want to go with JuliaFormatter.jl before giving actual review, but here I also want to comment a bit about configurations here:
Actually with [the `.JuliaFormatter.toml` configuration file](https://domluna.github.io/JuliaFormatter.jl/dev/config/), we don't need to handle by ourselves -- all what we need to do is to search for a configuration file and pass those configurations there like [Juno does](https://github.com/JunoLab/Atom.jl/blob/75eebb773ec222368801968ff5d6ae1e083c4fbf/src/formatter.jl#L11).
Then we get free from tasks to sync with these frontend configuration handling and actual formatter options (in other word we won't need something like [this PR](https://github.com/julia-vscode/LanguageServer.jl/pull/756) anymore.

pfitzseb: Yup, a config file is also how other prominent linters (e.g. ESLint) work and much more sensible than putting a bunch of options into every editor's UI.

non-Jedi: I think it's still valuable to provide configuration at the editor level so that formatting files always works as the user expects without having to create a `.JuliaFormatter.toml` file for one-off scripts (if e.g. the user prefers 2-space tabs as used in the DifferentialEquations codebase). I agree that we should hook this up so that a `.JuliaFormatter.toml` file overrides language server configuration.

pfitzseb: Yes, but imho that should only happen for "common" options like indentation (i.e. options that are shared by all languages).

domluna:
> Thanks for chiming in @domluna, if you could point out which methods would need to be defined for a custom style in order to prevent inserting line breaks (since that was a point of concern in [julia-vscode/DocumentFormat.jl#70](https://github.com/julia-vscode/DocumentFormat.jl/issues/70)), that would probably be helpful for this discussion. Or if it's more or less all of them, that's also valuable information. Is it as simple as defining `nest_if_over_margin!` to be a no-op?

Sort of, the main thing is to make check whether the text is on the same line in the source file and set `join_lines` based on that.

https://github.com/domluna/JuliaFormatter.jl/blob/master/src/styles/yas.jl#L72

Also don't insert placeholder nodes, if those aren't there the nesting stage is a noop.

Of course there should be still be indentation of some sort, the YAS style indentation seems to be the closest to DocumentFormat but that could be explored.

domluna:
I have a WIP DocumentFormat-esque style https://github.com/domluna/JuliaFormatter.jl/pull/256

ZacLN:
Thanks @non-Jedi for the really helpful issue. Broadly I'm in support of this, going through your list:

(I'll add I've not actually used JF so please forgive any misconceptions)

1. Missing features. I assume JF is far ahead DF as is. My only question - Is JF currently able to apply formatting on the fly (i.e. on-keystroke) or is it structured so that this would be possible (and performant)? Leaving that option open was (I think) one of the drivers when designing DF. 
2. This is my only concern. I don't like aggressive formatting, other do - there need to be options. I'd also suggest that they should fit into the vscode configuration story (including the option  within vscode to point to whatever format of config file JF uses).
3. Not concernd about this as long as 2. and 5. are resolved in a satisfactory manner. I minimally maintain DF and really don't have much interest in it.
4. I don't have as a strong feeling on this as others but wouldn't seem to be a deal-breaker to me.
5. I agree with the points made by David on the other thread, but as a risk it would seem to be manageable. A bit dependent on whether @domluna is interested/able to commit a little bit of time to it.

domluna:
> My only question - Is JF currently able to apply formatting on the fly (i.e. on-keystroke) or is it structured so that this would be possible (and performant)? Leaving that option open was (I think) one of the drivers when designing DF.

The most common use of JF is formatting the entire file on save. You can pass arbitrary julia code and it'll format it but without the context of the entire file this might not be ideal. However, this might be sufficient for this use case but I'm not familiar with the API requirements.

Timings formatting for a julia file read into REPL `s`:

```
julia> @btime DocumentFormat.format($s);
  18.235 ms (54518 allocations: 4.20 MiB)

julia> @btime JuliaFormatter.format_text($s);
  3.088 ms (47567 allocations: 2.85 MiB)
```

Timings for a small piece of text, I imagine this would be a touch type use case (?)

```
julia> str = "foo(argument1,                           argument2,                                                    argument3            )"
"foo(argument1,                           argument2,                                                    argument3            )"

julia> @btime DocumentFormat.format($str);
  74.314 μs (256 allocations: 15.69 KiB)

julia> @btime JuliaFormatter.format_text($str);
  8.825 μs (230 allocations: 16.06 KiB)
```

ZacLN:
That looks good (in terms of being much faster) - I was kind of asking whether it is structured so that, for example, given some small node of a file's CST you could quickly get the context necessary to format that node without also formatting the whole file

EricForgy:
I've never used JuliaFormatter, but this PR makes sense to me. I came across it because I wanted no spaces with `TypeVar` definitions.

>  Lack of julia-vscode governance.

Why not just move JuliaFormatter to an org? Maybe here?

domluna:
> That looks good (in terms of being much faster) - I was kind of asking whether it is structured so that, for example, given some small node of a file's CST you could quickly get the context necessary to format that node without also formatting the whole file

Partial formatting would probably be what you're looking for. This would still operator on the entire file (to figure out indentation and what not) but the output would only differ for the lines specified. So the lines that you don't want formatted would essentially be have formatting disabled https://domluna.github.io/JuliaFormatter.jl/stable/skipping_formatting/

singularitti:
Don't want to diverge here but I would love to see VSCode uses JF. I currently maintain [Julia Formatter](https://marketplace.visualstudio.com/items?itemName=singularitti.vscode-julia-formatter) and it has a major issue: speed. The reason is I need to start and stop a Julia process for each file to be formatted. This formatter is the only one I have used to be waiting until it's done. It would be faster, I think, if you have a running Julia process all the time and have JF precompiled.

aviatesk:
> That looks good (in terms of being much faster) - I was kind of asking whether it is structured so that, for example, given some small node of a file's CST you could quickly get the context necessary to format that node without also formatting the whole file

AFAICT, JuliaFormatter works as far as an input text represents a valid Julia expressions and is parsable by CSTParser (am I right, @domluna ?).
So as an editor integration, we can support "selection-based formatting", which I think it covers @ZacLN 's concern (as a result, Juno supports 1.) whole document formatting 2.) selection-based formatting, and  3.) on-save (whole-document) formatting).

As another note, I found DocumentFormat works even if a given text doesn't represent "valid Julia expression", and I guess this is the main difference in term of how formatting works in both pkgs.

domluna:
> AFAICT, JuliaFormatter works as far as an input text represents a valid Julia expressions and is parsable by CSTParser (am I right, @domluna ?)

yeah that's right

non-Jedi:
@ZacLN for mitigating point 5, would it be desirable to have usage of JuliaFormatter be opt-in via a configuration option for a release or two? Since DocumentFormat is really only called in one place, this wouldn't be difficult to accomplish.

jlperla:
Not to be "that guy" who bumps PRs without contributing code, but...

Any hope on getting these sorts of features into vscode /either directly in the existing formatter or adopting this one?

samuela:
https://github.com/julia-vscode/julia-vscode/issues/1725#issuecomment-719248932 appears to indicate that this has actually been closed/merged. Is that the case?

pfitzseb:
No. We still want to get this in. A common dependency has changed a lot though, so this is on hold until JuliaFormatter is compatible with CSTParser@3.

domluna:
> No. We still want to get this in. A common dependency has changed a lot though, so this is on hold until JuliaFormatter is compatible with CSTParser@3.

I'll get this upgraded this week

@ZacLN was the generator thing we discussed on slack a bug or is that the new way it outputs now?

ZacLN:
> @ZacLN was the generator thing we discussed on slack a bug or is that the new way it outputs now?

Def a bug, it's the correct way for the `EXPR` to be stored so as to match `Expr` but I need to find a workaround for `getindex(x::EXPR, i)` so as to make it iterate in display order

ZacLN:
Fix for :flatten is at https://github.com/julia-vscode/CSTParser.jl/pull/259

domluna:
Ok things seems to be working on CSTParser#master. There might more bugs not caught by the unit or integration tests but I won't know until other people use it

src/formatoptions.jl
davidanthoff: How about this: we make the VS Code config settings scope `application`, so they can _only_ be configured in user settings, but not as workspace or folder settings. I think what we should try to avoid is to have two different ways to configure folder specific settings, and I think the separate `.JuliaFormatter.toml` story then makes more sense than a VS Code specific configuration setting with `window` scope.

I think in terms of implementations this would mean that a `.JuliaFormatter.toml` would always take precedence (if it exists) over any config settings that the LS receives from the client, right?

What is a lot less clear to me is which config settings we should have in the client... Minimally I guess the option to specify a style? But I don't really see a good reason _not_ to allow all the other settings to be specified as well...

domluna:
What's the blocker for this atm?

davidanthoff:
I think the main blocker is that a) we're still not done with the major other work happening in LS land, and I'd rather have us finish us those parts first before we start a new big change, but more generally b) that no one had time for much lately :) But maybe we should just say that after Juliacon we make this here a priority.

pfitzseb:
JuliaFormatter will also currently bring in 7 new dependencies, which isn't too great (good news is that there don't appear to be any binary dependencies, so distribution should just work).

pfitzseb:
Ok, I have no idea what GitHub (or I) did with the commit history here. Will close this in favor of https://github.com/julia-vscode/LanguageServer.jl/pull/972.

PR #761, opened 2020-06-16
simplify kw completion snippet code

ZacLN:
in place of #740. Also fixes the case of missing completions when on a keyword.


src/requests/completions.jl
```diff
@@ -63,106 +65,50 @@ function latex_completions(doc, offset, partial, CIs)
...
+    "function" => "function \$0(\$1)\nend",
+    "global" => "global ",
+    "if" => "if \$0\nend",
+    "import" => "import",
+    "let" => "let \$0\nend",
+    "local" => "local ",
+    "macro" => "macro \$0(\$1)\nend",
+    "module" => "module \$0\nend",
+    "mutable" => "mutable struct \$0\nend",
+    "outer" => "outer ",
```
non-Jedi, line 102: Did `outer` have some sort of special meaning before 1.0? Or is there a form using `outer` that I'm not aware of?

```
❯ julia --lisp
;  _
; |_ _ _ |_ _ |  . _ _
; | (-||||_(_)|__|_)|_)
;-------------------|----------------------------------------------------------

> reserved-words
(begin while if for try return break continue function macro quote let local
 global const do struct module baremodule using import export end else elseif
 catch finally true false)
```

src/requests/completions.jl
```diff
@@ -63,106 +65,50 @@ function latex_completions(doc, offset, partial, CIs)
...
+    "catch" => "catch",
+    "const" => "const ",
+    "continue" => "continue",
+    "do" => "do \$0\nend",
+    "else" => "else",
+    "elseif" => "elseif ",
+    "end" => "end",
+    "export" => "export ",
+    "finally" => "finally",
+    "for" => "for \$0 in \$1\nend",
```
non-Jedi, line 92:
```suggestion
    "for" => "for \$0 in \$1\n    \$2\nend",
```

src/requests/completions.jl
```diff
@@ -63,106 +65,50 @@ function latex_completions(doc, offset, partial, CIs)
...
+    "function" => "function \$0(\$1)\nend",
```
non-Jedi, line 93:
```suggestion
    "function" => "function \$0(\$1)\n    \$2\nend",
```

src/requests/completions.jl
```diff
@@ -63,106 +65,50 @@ function latex_completions(doc, offset, partial, CIs)
...
 
+const snippet_completions = Dict{String,String}(
+    "abstract" => "abstract type \$0 end",
+    "baremodule" => "baremodule \$0\nend",
+    "begin" => "begin\$0\nend", 
+    "break" => "break",
+    "catch" => "catch",
+    "const" => "const ",
+    "continue" => "continue",
+    "do" => "do \$0\nend",
```
non-Jedi, line 86: `do` blocks can also have a zero-argument form.

src/requests/completions.jl
ZacLN: It soft-scopes (is that a term?) the iterator of a loop to the parent (local) scope such that:
```julia
let
    i = 0
    for outer i = 1:10
    end
    @assert i == 10
end
```

src/requests/completions.jl
non-Jedi: Didn't realize that [the editor visits the tab stops in reverse order](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#snippet_syntax). In general, we should number the tab-stops so they advance rather than jump around.

```suggestion
    "function" => "function \$2(\$1)\n    \$0\nend",
```

src/requests/completions.jl
```diff
@@ -63,106 +65,50 @@ function latex_completions(doc, offset, partial, CIs)
...
+    "else" => "else",
+    "elseif" => "elseif ",
+    "end" => "end",
+    "export" => "export ",
+    "finally" => "finally",
+    "for" => "for \$0 in \$1\n    \$2\nend",
```
non-Jedi, line 92: Slightly fancier:

```suggestion
    "for" => "for \$3\${2|=, in |}\$1\n    \$0\nend",
```

src/requests/completions.jl
ZacLN: Ah, I'd forgotten that - I thought there were errors in the original code

src/requests/completions.jl
ZacLN: I'm going to adjust the PR to provide the most barebones implementation I can think of. 

Personally I would prefer a single placeholder for each entry as completions (etc) don't seem to be enabled when jumping between the tab stops. Thoughts on that?

ZacLN: I also don't think `itr = Something` reads particularly well but that could just be me!

non-Jedi: I think completions not working when in the midst of filling a snippet might be just a vscode issue. That said, these kind of snippets aren't super useful compared to e.g. filling out function argument placeholders in the future if we generate snippets for specific functions/methods. I'm fine with leaving them *relatively* simple.

PR #765, opened 2020-06-18
nothing -> missing in WorkspaceEdit

### LINKED ISSUE #763: Don't send "changes": null in WorkspaceEdit on workspace/applyEdit request
The server sends `"changes": null` within WorkspaceEdit in the `workspace/applyEdit` request (after receiving a `workspace/executeCommand`). But according to the specs for [WorkspaceEdit](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspaceEdit), `changes` must not be `null`. It can just be omitted instead.

#665 doesn't seem to have resolved that.

ZacLN:
closes #763

This is because `WorkspaceEdit` is mis-specified within our version of the spec, what does: 
`ts { [uri: DocumentUri]: TextEdit[]; };` mean? 

Is it a vector of pairs `(DocumentUri, Vector{TextEdit})`?

### DIFF
--- a/src/requests/actions.jl
+++ b/src/requests/actions.jl
@@ -115,7 +115,7 @@ function explicitly_import_used_variables(x::EXPR, server, conn)
         return
     end
 
-    JSONRPC.send(conn, workspace_applyEdit_request_type, ApplyWorkspaceEditParams(missing, WorkspaceEdit(nothing, collect(values(tdes)))))
+    JSONRPC.send(conn, workspace_applyEdit_request_type, ApplyWorkspaceEditParams(missing, WorkspaceEdit(missing, collect(values(tdes)))))
 end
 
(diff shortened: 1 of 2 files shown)


davidanthoff:
> what does: ts `{ [uri: DocumentUri]: TextEdit[]; };` mean?

Maybe `Dict{String, Vector{TextEdit}}`? But I'm not sure ;)

davidanthoff:
How about we try
```julia
changes::Union{Dict{DocumentUri, Vector{TextEdit}},Missing}
```
Is there a reason we aren't trying 
```julia
documentChanges::Union{
    Missing,
    Vector{TextDocumentEdit},
    Vector{Union{TextDocumentEdit,CreateFile,RenameFile,DeleteFile}}
}
```
as well?

PR #772, opened 2020-06-19
Introduce runserver function

### LINKED ISSUE #748: Don't require `env_path` to be passed to `LanguageServerInstance`
When using `LanguageServerInstance` in `neovim`, the default configuration currently starts a process for `LanguageServer.jl` in the current working directory:

https://github.com/neovim/nvim-lsp/blob/94cce7448f34f172492d98ff6efe10986f73eb9e/lua/nvim_lsp/julials.lua#L14

```
        project_path = pwd()
		...
        server = LanguageServer.LanguageServerInstance(stdin, stdout, project_path, depot_path);
```

When using vim / neovim, users are going to possibly going to do this:

```
cd Package.jl/src/module1/
nvim filename.jl
```

This means that LanguageServer will be started in the `Package.jl/src/module1` folder instead of at `Package.jl` where the `Project.toml` file is located.

Additionally, one can imagine a Package.jl that has some tests that depend on external packages. In this case, Package.jl/tests/extras may contain a Project.toml file that specifies some dependencies for running `test1.jl`.

```
cd Package.jl/tests/extras/
nvim test1.jl
```

LanguageServer.jl/SymbolServer.jl will fail to find references / lint correctly in this case.

Would it be possible to have `project_path` recursively search up from the path provided till it finds a `Project.toml`?

non-Jedi:
runserver will attempt to pick a reasonable environment to run the
language server against.

Closes #748.

We still need to address running the language server against a project
with no Manifest.

Long term we also probably want to move away from the `env_path` being passed into the language server, but this at least makes things easier on clients until then.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,17 +21,22 @@ make use of the Julia Language Server for various code editing features:
 ## Installation and Usage
 **Documentation**: [![Dev](https://img.shields.io/badge/docs-dev-blue.svg)](https://www.julia-vscode.org/LanguageServer.jl/dev)
 
+To install LanguageServer.jl into the current environment:
+
 ```julia
 using Pkg
 Pkg.add("LanguageServer")
 ```
 
-Instantiate an instance of the language server with
-`LanguageServerInstance` and `run` it:
+To run an instance of LanguageServer.jl on `env_path`, you can run
+Julia as follows:
 
-```julia
-using LanguageServer
-
-server = LanguageServerInstance(stdin, stdout, "/path/to/environment")
-run(server)
+```sh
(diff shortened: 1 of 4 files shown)


non-Jedi:
@puremourning does this work better for you a a stopgap solution (see the updated README)? Will YouCompleteMe start the language server with `pwd` in the directory of the file being edited?

davidanthoff:
> We still need to address running the language server against a project with no Manifest.

Yes, that will be part of the "big" env handling rework that will come at some point (hopefully soon).

PR #776, opened 2020-06-22
disable linting in test folders

ZacLN:


src/utilities.jl
```diff
@@ -331,3 +331,49 @@ function op_resolve_up_scopes(x, mn, scope, server)
 end
 
 maybe_lookup(x, server) = x isa SymbolServer.VarRef ? SymbolServer._lookup(x, getsymbolserver(server), true) : x # TODO: needs to go to SymbolServer
+
+function is_in_test_dir_of_package(path::String)
+    try # Safe failure - attempts to read disc.
+        if VERSION < v"1.1"
+            return false
+        else
```
aviatesk, line 339: Maybe you just forgot to remove this check ?

pfitzseb:
Does this disable _all_ linting or only "global" linting? It would be nice to still have some for of checking in local files, for example.

src/utilities.jl
ZacLN: Yep!

ZacLN:
I don't understand what you mean. This disables linting for _files_ located within a "test" sub-directory of a folder that is a Julia package folder

pfitzseb:
Right, but it would be cool if it didn't disable _all_ linting -- some features are local to the current file (and maybe `include`s), but don't need to care about packages at all. E.g. the analysis of which function arguments are used etc. So if possible we should keep that kind of stuff enabled.

davidanthoff:
@ZacLN and @pfitzseb what do we do about this as a next step?

aviatesk:
Is it easy to do "basic lintings" (i.e. syntax-errors and such, those could be done by per-expression or per-file) separately from "more complex linting" (like warning on undef name, which would be done by per-file or per-package level) ?
If not, I think we want to mergd this as is and do the separation first in another PR ?

ZacLN:
This now disables lint checks that are dependent on having an accurate representation of the package environment. Not a particularly elegant approach, I need to start categorising LintCodes.

pfitzseb:
Nice, this is exactly what I had in mind! Can always fix up the implementation at a later point

PR #778, opened 2020-06-29
check docversion in module/blockrange requests

pfitzseb:


src/protocol/document.jl
```diff
@@ -61,6 +61,12 @@ end
     position::Position
 end
 
+@dict_readable struct VersionedTextDocumentPositionParams
+    textDocument::TextDocumentIdentifier
+    version::Union{Int,Nothing}
```
davidanthoff, line 66: Should this really be a `Union`? Wouldn't it make more sense to just require `Int` here all the time?

src/requests/features.jl
```diff
@@ -300,14 +300,23 @@ function _binding_kind(b, server)
...
+            x = get_expr(getcst(doc), offset)
+            if x isa EXPR
+                scope = StaticLint.retrieve_scope(x)
+                if scope !== nothing
+                    return get_module_of(scope)
+                end
+            end
+        else
+            @warn "version mismatch in getModuleAt for $(uri): internal $(doc._version), in vscode: $(params.version)"
+            return nothing
```
davidanthoff, line 319: We can't return a `nothing` here, that is not a valid return value according to the request definition (I should actually add a check to JSONRPC.jl that throws an error if a handler does this kind of thing).

If you want to return an error condition, just return a `JSONRPCError` (don't throw it, just return it).

We should also definitely make sure we record this situation in the telemetry, just to understand whether version sync story _is_ actually the root case of this problem or not. The brute force version for now would be to define another one of our custom error types and throw an error here and crash for now, or you could also send a trace.

src/requests/features.jl
```diff
@@ -300,14 +300,23 @@ function _binding_kind(b, server)
...
-    offset = get_offset(doc, params.position)
-    x = get_expr(getcst(doc), offset)
-    if x isa EXPR
-        scope = StaticLint.retrieve_scope(x)
-        if scope !== nothing
-            return get_module_of(scope)
+function julia_getModuleAt_request(params::VersionedTextDocumentPositionParams, server::LanguageServerInstance, conn)
+    uri = URI2(params.textDocument.uri)
+
+    if hasdocument(server, uri)
```
davidanthoff, line 306: I don't think we should have this check here. If seems to me that we have an error if the doc doesn't exist in the LS, so we should just crash in that situation to make sure we learn of the problem and can then fix it. I would just remove this `if` clause here, so that `getdocument` crashes if the doc doesn't exist.

src/requests/misc.jl
```diff
@@ -5,8 +5,19 @@ end
 function setTraceNotification_notification(params, server::LanguageServerInstance, conn)
 end
 
-function julia_getCurrentBlockRange_request(tdpp::TextDocumentPositionParams, server::LanguageServerInstance, conn)
-    doc = getdocument(server, URI2(tdpp.textDocument.uri))
+function julia_getCurrentBlockRange_request(tdpp::VersionedTextDocumentPositionParams, server::LanguageServerInstance, conn)
+    fallback = (Position(0, 0), Position(0, 0), tdpp.position)
+    uri = URI2(tdpp.textDocument.uri)
+
+    hasdocument(server, uri) || return fallback
```
davidanthoff, line 12: Same here, we should crash if the `doc` doesn't exist, something must have gone really wrong in that situation.

src/requests/misc.jl
```diff
@@ -5,8 +5,19 @@ end
...
+
+    doc = getdocument(server, uri)
+
+    if doc._version !== tdpp.version
+        @warn "version mismatch in getCurrentBlockRange for $(uri): internal $(doc._version), in vscode: $(tdpp.version)"
+        return fallback
```
davidanthoff, line 18: I don't think we should return the `fallback` here, but an error condition.

src/protocol/document.jl
pfitzseb: Yes, I just copy-pasted this from the "official" `TextDocumentPositionParams` above.

src/requests/features.jl
pfitzseb: Hm, I don't think that's a good idea -- we can't ever guarantee in which order our custom requests are interleaved with the LS internals, right?

davidanthoff: Yeah, I guess that is true... We should probably return a `JSONRPCError` then and handle that situation in the client?

ZacLN:
Sorry, late to the party on this. Could we move this code from the `protocol` folder and indicate that it's not part of the LSP?

pfitzseb:
Yeah, I can do that.

Issue #747, opened 2020-06-10
`CompletionItem` always defaults to being a `Snippet`


kdheepak:
[The specification](https://microsoft.github.io/language-server-protocol/specification) allows `InsertTextFormat` to be plain text or a snippet.

```
/**
 * Defines whether the insert text in a completion item should be interpreted as
 * plain text or a snippet.
 */
export namespace InsertTextFormat {
	/**
	 * The primary text to be inserted is treated as a plain string.
	 */
	export const PlainText = 1;

	/**
	 * The primary text to be inserted is treated as a snippet.
	 *
	 * A snippet can define tab stops and placeholders with `$1`, `$2`
	 * and `${3:foo}`. `$0` defines the final tab stop, it defaults to
	 * the end of the snippet. Placeholders with equal identifiers are linked,
	 * that is typing in one will update others too.
	 */
	export const Snippet = 2;
}
```

Currently in `LanguageServer.jl`, creating an instance of `CompletionItem` always defaults to being a `Snippet`. 

https://github.com/julia-vscode/LanguageServer.jl/blob/f9edb8091b067fc296170333a67a037af20387c4/src/protocol/completion.jl#L106

The 11th argument is hardcoded as `2`.

https://github.com/julia-vscode/LanguageServer.jl/blob/f9edb8091b067fc296170333a67a037af20387c4/src/protocol/completion.jl#L99

This seems to break completion in `neovim`. 

Can `LanguageServer.jl` be updated to use `PlainText` by default and use `Snippet` explicitly whenever required?

PR #780, opened 2020-06-29
return completions as plaintext by default

ZacLN:


non-Jedi:
I'm assuming this closes #747?

PR #788, opened 2020-07-02
move julia specific LS extensions to seperate dir

pfitzseb:
As requested by @ZacLN in https://github.com/julia-vscode/LanguageServer.jl/pull/778#issuecomment-652434721.


ZacLN:
Thanks!

pfitzseb:
At some point I'd like to make the same split for the actual request handlers and utility functions, but that's much more of a big refactor and we probably shouldn't do that now.

ZacLN:
Yep agreed

PR #793, opened 2020-07-06
Use MD for julia_getDocAt_request, remove all VS Code specifics

davidanthoff:
The corresponding PR on the VS Code side is coming in a minute, I'll explain there.


davidanthoff:
I'm merging this now because it hopefully doesn't affect anything that we ship right now and it will make it a lot easier to work on the doc branches over in julia-vscode.

PR #794, opened 2020-07-11
Fix a race condition for progress tokens

davidanthoff:
Fixes a race condition that came in via crash reporting.

The problem is that during the `JSONRPC.send` call other tasks can set `server.current_symserver_progress_token` to `nothing`, and then we crash. Here we use a local variable now to for `send` calls, which ensures that we can't get into that state.

### DIFF
--- a/src/languageserverinstance.jl
+++ b/src/languageserverinstance.jl
@@ -137,13 +137,14 @@ end
 
 function create_symserver_progress_ui(server)
     if server.clientcapability_window_workdoneprogress
-        server.current_symserver_progress_token = string(uuid4())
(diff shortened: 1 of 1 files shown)


davidanthoff:
I'm merging this even though the CI fails. The CI failure is already on `master`, so this is not causing it, and this PR here is definitely solving crash reports we are getting.

I don't understand how we ended up with a broken `master`, though.... The only explanation I have is that I tagged a new JSONRPC.jl release, so maybe we ran CI here with an older version for quite a while already and something was broken there. I'll try to fix that first thing tomorrow.

PR #800, opened 2020-07-15
Escape from get_func_hover loop

ZacLN:
There is a cycle in the links (`.prev`/`.next`) between StaticLint.Bindings. The root cause is within StaticLint but there is no reason to allow it to crash the language server. If we have done a complete circuit here (i.e. a loop is detected) then we have all the information we need and can `return`.

As this is being hit quite a lot, this approach seems best for the user while the StaticLint bug is fixed. Thoughts?

### DIFF
--- a/src/requests/features.jl
+++ b/src/requests/features.jl
@@ -1,6 +1,6 @@
 function get_signatures(b, tls, sigs, server, visited = nothing) end # Fallback
 
-function get_signatures(b::StaticLint.Binding, tls::StaticLint.Scope, sigs::Vector{SignatureInformation}, server, visited = Base.IdSet{StaticLint.Binding}())
+function get_signatures(b::StaticLint.Binding, tls::StaticLint.Scope, sigs::Vector{SignatureInformation}, server, visited = StaticLint.Binding[])
     if b in visited                                      # TODO: remove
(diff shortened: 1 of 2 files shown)


aviatesk:
I'm in favor of this -- as far as I understand binding traversal in LanguageServer.jl itself seems to be correct, so I think there won't be anything we can get more from crash report for these lines of code.

src/requests/features.jl
```diff
@@ -1,6 +1,6 @@
 function get_signatures(b, tls, sigs, server, visited = nothing) end # Fallback
 
-function get_signatures(b::StaticLint.Binding, tls::StaticLint.Scope, sigs::Vector{SignatureInformation}, server, visited = Base.IdSet{StaticLint.Binding}())
+function get_signatures(b::StaticLint.Binding, tls::StaticLint.Scope, sigs::Vector{SignatureInformation}, server, visited = StaticLint.Binding[])
```
aviatesk, line 3: What's the reason for switching to `Vector` ? 
Well, could I also ask why you used `Base.IdSet` before instead of usual `Set` ?

ZacLN: Using IdSet was a mistake, I had been using them in SymbolServer and forgot it was quite costly

Issue #768, opened 2020-06-19
get_offset crashed


davidanthoff:
From [crash reporting](https://portal.azure.com/#blade/AppInsightsExtension/DetailsV2Blade/ComponentId/%7B%22SubscriptionId%22%3A%226803c4ed-bcb4-41d4-bceb-7faf5e7a3469%22%2C%22ResourceGroup%22%3A%22Default%22%2C%22Name%22%3A%22julia-vscode%22%2C%22LinkedApplicationType%22%3A0%2C%22ResourceId%22%3A%22%2Fsubscriptions%2F6803c4ed-bcb4-41d4-bceb-7faf5e7a3469%2FresourceGroups%2FDefault%2Fproviders%2Fmicrosoft.insights%2Fcomponents%2Fjulia-vscode%22%2C%22ResourceType%22%3A%22microsoft.insights%2Fcomponents%22%2C%22IsAzureFirst%22%3Afalse%7D/DataModel/%7B%22eventId%22%3A%22727d6651-b1b1-11ea-8021-9527106bc9e8%22%2C%22timestamp%22%3A%222020-06-18T22%3A17%3A07.472Z%22%2C%22cacheId%22%3A%22668a0b57-2106-4b27-95f0-eece2e7ba3d1%22%2C%22eventTable%22%3A%22exceptions%22%7D).

Full error message there, not here, because it reveals too much about the client system.

Stack trace:
```
LanguageServer.LSOffsetError:
   at get_offset(::LanguageServer.Document, ::Int64, ::Int64) (./julialang.language-julia-0.16.7/scripts/packages/LanguageServer/src/document.jl93)
   at get_offset (./julialang.language-julia-0.16.7/scripts/packages/LanguageServer/src/document.jl96)
   at julia_getModuleAt_request(::LanguageServer.TextDocumentPositionParams, ::LanguageServerInstance, ::JSONRPC.JSONRPCEndpoint) (./julialang.language-julia-0.16.7/scripts/packages/LanguageServer/src/requests/features.jl305)
   at (::LanguageServer.var"#16#47"{LanguageServerInstance})(::JSONRPC.JSONRPCEndpoint, ::LanguageServer.TextDocumentPositionParams) (./julialang.language-julia-0.16.7/scripts/packages/LanguageServer/src/languageserverinstance.jl278)
[...]

PR #807, opened 2020-07-22
Fix an offset bug

davidanthoff:
Next attempt to fix https://github.com/julia-vscode/LanguageServer.jl/issues/768.

This is kind of embarrassing because I completely forgot about the whole `offset` vs `offset2` thing in the code base... Here is the background story: we had a lot of text sync errors in the fall, so I rewrote the whole text sync logic, largely just copying the algorithm from the original MS node implementation. One thing I was never able to square was the original `get_offset` implementation we had and one that followed the MS logic (that is to say, I didn't really understand our implementation :) ). In the end I added the MS based implementation as `get_offset2` and used that for all the text sync stuff. That got rid of all the text sync bugs. I didn't remove the original `get_offset` implementation because I didn't fully understand the ways it was used in various other parts of the LS, and was worried that my new implementation actually used a slightly different definition than our original implementation. So, I left it in there and some of the request handlers still use it today.

So one simple explanation for our problems here is that the old `get_offset` implementation might just be buggy. The fact that swapping it out for the text sync stuff fixed all the problems around that suggests so :) I don't really know why I never followed up on this and tried to generally use `get_offset2` everywhere in the code base, I think I simply forgot...

So this just swaps things out and uses the `get_offset2` implementation here. Maybe that will just fix things :)

### DIFF
--- a/src/requests/features.jl
+++ b/src/requests/features.jl
@@ -349,7 +349,7 @@ function julia_getModuleAt_request(params::VersionedTextDocumentPositionParams, 
     if hasdocument(server, uri)
         doc = getdocument(server, uri)
         if doc._version == params.version
-            offset = get_offset(doc, params.position)
+            offset = get_offset2(doc, params.position.line, params.position.character)
             x = get_expr(getcst(doc), offset)
             if x isa EXPR
                 scope = StaticLint.retrieve_scope(x)


davidanthoff:
Note to myself: if this works, we should make the same change to hover and definition requests, where we are also still seeing offset errors.

pfitzseb:
I did spend an hour or two playing around with `get_offset` and it seemed correct to me, but let's see how this pans out.

davidanthoff:
Yeah, I was never able to pin down or identify a bug in the original `get_offset`, but not using it fixed the sync issues ;) I think if there is a problem, it must be some weird corner case...

ZacLN:
If this is unsuccessful and there are still questions as to whether we the root issue is with our incremental document updating we have the option of moving back to 'full text updating'. This would have performance penalties but should give us certainty over the problem..

davidanthoff:
I'm fairly positive that the text sync is correct now, otherwise we should see tons of https://github.com/julia-vscode/LanguageServer.jl/blob/fa14ca17b8899e0578f378ed24a38b5b96204ffe/src/requests/textdocument.jl#L54, right? And I think we haven't seen that at all for a very long time.

pfitzseb:
We did just get e.g. 
```
get_offset crashed. More diagnostics:
line=176
character=0
position(io)=0
line_offsets='[0]'
text=''

original_error=ERROR: BoundsError: attempt to access 1-element Array{Int64,1} at index [177]
```
on (still on 0.17.7). That does kinda smell like a sync error unrelated to the document version stuff. Maybe there's an issue where text edits aren't applied properly in very specific circumstances or something...

pfitzseb:
We haven't gotten a report about this on 0.17.8 yet...

aviatesk:
I think we have just one crash report (7 hours ago) ?

pfitzseb:
No, we got a few more that I didn't realize were related: 
https://github.com/julia-vscode/LanguageServer.jl/blob/5549182d9435059383f1d5f6c5127378e6d8b89e/src/document.jl#L104
isn't as easy to make out as `LSOffsetError`. Ok, so this didn't fix it either.

aviatesk:
agh, true... #809

ZacLN:
So my ideal way forward from here would be to revert to full document syncing on an insiders build release to confirm that is the issue. Then I would like to add some  inspection code that checks the full text between client and server after each update, storing text change events (obfuscated) that then are sent as part of the error when we get failures such as the above. This allows us to recreate the failed update.

Alternatively we could do more reviewing of the incremental document update code or chuck it out and start again from scratch.

davidanthoff:
I do have the impression, though, that this PR here reduced the number of users affected by quite a lot? Which makes all of this even more weird, because if this is caused by a sync error, then this PR here should _not_ have made any difference at all...

Agreed that maybe we should try to the non incremental sync for a while. I don't know whether we have enough crash reports for this particular problem on insider, though, to really get a good sense what is going on. Could we switch the release channel over to full sync for a few days at some point and see what happens?

I also think that we probably just have to postpone further attempts to fix this until after Juliacon? I'm a bit worried that any further diagnostic work we might do could introduce other unintended side effects...

PR #811, opened 2020-08-01
fix latex completions with multiple cursors

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/issues/1334.


aviatesk:
I think we can merge this as is; this looks and feels good as far as I see. @ZacLN can leave the comment later :)

PR #813, opened 2020-08-05
disable incremental document updates

pfitzseb:
Might or might not help with the `get_offset` issues...

### DIFF
--- a/src/requests/init.jl
+++ b/src/requests/init.jl
@@ -1,6 +1,6 @@
 const serverCapabilities = ServerCapabilities(
     TextDocumentSyncOptions(true,
-    TextDocumentSyncKinds.Incremental,
+    TextDocumentSyncKinds.Full,
     false,
     false,
     SaveOptions(true)),


aviatesk:
😭

davidanthoff:
We should probably try this, but on the other hand, shouldn't we see https://github.com/julia-vscode/LanguageServer.jl/blob/d55931ee71aef0d62b297d60f25176773edcc2a2/src/requests/textdocument.jl#L54 in the crash reporting if there really still was a text sync error? I guess, though, we could be in a scenario where we have a text sync error, and then the LS always crashes _before_ it reaches that piece of code... So lets merge this here and try it for a few days, and then undo quickly, hopefully :)

pfitzseb:
Well, if our incremental sync is subtly broken then that could explain the errors in `get_offset`. Those kinds of errors would still show the correct document versions, but the underlying document on the LS side would be corrupt.
That said, I don't think this will actually fix/change anything ;)

davidanthoff:
> but the underlying document on the LS side would be corrupt.

Yeah, but the link I posted above is a check on the _full_ document content, not the version. So every time a user saves a file, we completely check whether our copy of the text matches the version of the text on the VS Code side. I would have thought that this should pick up any sync issues we might have... But I'll merge now and we'll try this out :)

PR #816, opened 2020-08-12
add ^ to completion triggers

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/issues/1595.

### DIFF
--- a/src/requests/init.jl
+++ b/src/requests/init.jl
@@ -4,7 +4,7 @@ const serverCapabilities = ServerCapabilities(
     false,
     false,
     SaveOptions(true)),
-    CompletionOptions(false, [".", "@", "\""], missing),
+    CompletionOptions(false, [".", "@", "\"", "^"], missing),
     true,
     SignatureHelpOptions(["(", ","], missing),
     false,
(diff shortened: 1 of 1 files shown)


non-Jedi:
Why is this needed? Linked vscode bug doesn't make clear why the language server should consider ^ a completion trigger.

pfitzseb:
It's not really a VSCode bug. 
Frontends will usually (as per the LSP docs, at least) trigger autocompletions for valid identifiers only, which e.g. `\^2` isn't. So with this we trigger completions on `^` anyways.

non-Jedi:
Why isn't `\` enough since it's already in the trigger list? That's the part that seems like working around something. So you trigger completion when `\` is typed; why does completion need to be triggered again on the following `^`?

non-Jedi:
Actually I think I see the reasoning here. If you've got a bare backslash already in your document and later come back and type `^` after it, that should still trigger completion. So  we should additionally add any other characters that aren't part of identifier syntax but can be part of LaTeX codes. Off the top of my head, `:` (`\:watermelon:`) and `/` (`\1/2`).

non-Jedi:
Scrolling through <https://docs.julialang.org/en/v1/manual/unicode-input/>, we also need:

- `+`
- `-`
- `=`
- `(`
- `)`

PR #834, opened 2020-09-06
Fix choose_env

pxl-th:
Both `Base.current_project` and `Base.load_path_expand` return a path to `Project.toml` file,
but `choose_env` needs to return a directory path containing that file, according to documentation.

This also causes lots of `Missing reference` errors right now, but with this fix everything is working fine.


davidanthoff (approved):
Great, thanks!

PR #839, opened 2020-10-10
feat: get_project_language_server

aminya:
Adds a function to get the project's language server

Analogous to https://github.com/julia-vscode/SymbolServer.jl/pull/190

# TODO 
Add tests


davidanthoff:
What is the motivation for this? This package is really meant to be used by editor integrations, and they will most likely need something that is a bit more robust than this, right?

In general I'd like to keep the API surface area of this package small, and ideally move things like this to an examples or contrib folder where we don't make guarantees that this is part of the official API.

PR #846, opened 2020-11-10
add "julia/fileSearchUp" request

aviatesk:


aviatesk:
for now we use JS to search up a file

PR #850, opened 2020-11-24
include link to new Kakoune wiki entry

koehlerson:
I've added a Wiki entry which describes a kakoune setup with julia lsp integration. This PR aims to link the new entry in the `README.md`


pfitzseb:
I think we'll end up moving the wiki to our normal docs at some point, but this is great to have for now.

pfitzseb (approved):

PR #851, opened 2020-11-29
throw source with errs

ZacLN:
Throws the source file contents that we're trying to parse when we hit a CSTInfiniteLoop. Can't get to the bottom of the loops without this, I don't think. A discussion is needed as to whether this is acceptable (beyond expectations of what users diagnostics send to us should include).

### DIFF
--- a/src/requests/textdocument.jl
+++ b/src/requests/textdocument.jl
@@ -227,7 +227,16 @@ function parse_all(doc::Document, server::LanguageServerInstance)
     if endswith(doc._uri, ".jmd")
         doc.cst, ps = parse_jmd(ps, get_text(doc))
     else
-        doc.cst, ps = CSTParser.parse(ps, true)
+        try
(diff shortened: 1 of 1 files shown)


pfitzseb:
Imho we shouldn't do this, especially not without notifying people. What do you think about running LanguageServer/StaticLint over the whole package ecosystem? That shouldn't take super long and might make the issue obvious.

ZacLN:
I feel the same as you - this is clearly bad - but I'm struggling to think of other ways to resolve this long standing issue. I am almost certain these errors are from parsing invalid syntax that comes about as code is typed. I've already run staticlint across the ecosystem at some point in the last year (found and fixed bugs but not this one).

Thoughts on what else we could do? I have a naive obfuscator in LS somewhere already, maybe putting some work/thinking into that would make this satisfactorily less intrusive?

PR #856, opened 2021-01-07
Cst compat

ZacLN:
Amendments for CSTParser changes (https://github.com/julia-vscode/CSTParser.jl/pull/190)


src/requests/textdocument.jl
```diff
@@ -68,130 +68,17 @@ end
 
 
 function textDocument_didChange_notification(params::DidChangeTextDocumentParams, server::LanguageServerInstance, conn)
+    t=  time()
```
pfitzseb, line 71:
```suggestion

```

src/runserver.jl
```diff
@@ -49,3 +49,10 @@ end
 choose_env() = something(get(ARGS, 1, nothing),          # 1. path passed explicitly
                          Base.current_project(pwd()),    # 2. parent project of pwd()
                          Base.load_path_expand("@v#.#")) # 3. default "global" env
+
+function choose()
+    maybe_dirname = x -> x !== nothing ? dirname(x) : nothing
+    something(get(ARGS, 1, nothing),                            # 1. path passed explicitlysomething(get(ARGS, 1, nothing))
```
pfitzseb, line 55:
```suggestion
    something(get(ARGS, 1, nothing),                            # 1. path passed explicitly
```

src/requests/init.jl
```diff
@@ -24,7 +24,7 @@ const serverCapabilities = ServerCapabilities(
     true,
     false,
     ExecuteCommandOptions(missing, collect(keys(LSActions))),
-    false,
+    true,
```
pfitzseb, line 27: This is `selectionRangeProvider`, right? Did we always support that? If so, why was it disabled?

pfitzseb (commented):
Let's turn incremental sync back on and see what happens, at least in the next insiders release? I kinda hope that all the code churn here and in CSTParser might've fixed the issues we had.

src/requests/init.jl
ZacLN: I must have been fiddling with something, will turn off

ZacLN:
@julia-vscode/core could someone with a windows computer try to get this passing? Debugging by pushing to CI runs is a nightmare.

The issue is with path handling on older Julia verions (<1.1 I think, possibly <1.3). Relevant code is https://github.com/julia-vscode/LanguageServer.jl/blob/master/src/utilities.jl#L286:L309 or there abouts

src/requests/init.jl
```diff
@@ -1,6 +1,6 @@
 const serverCapabilities = ServerCapabilities(
     TextDocumentSyncOptions(true,
-    TextDocumentSyncKinds.Full,
+    TextDocumentSyncKinds.Incremental,
```
pfitzseb, line 3: Let's try re-enabling this?

pfitzseb (approved):

PR #862, opened 2021-01-20
Incremental update for cst/semantic

ZacLN:
Partner to https://github.com/julia-vscode/StaticLint.jl/pull/243. 
Allows incremental updating of the cst/semantic info without relying on updating the text incrementally (which I think we suspect is problematic).


ZacLN:
Needs new versions for CSTParser and StaticLint but otherwise ready.

davidanthoff (approved):

PR #866, opened 2021-01-24
Update to new client lib version

davidanthoff:
The new client side lib version sends a new client id, and so our insider build is crashing at the moment. I'm going to merge this right away to fix that. Review after the fact would be great.

### DIFF
--- a/src/requests/workspace.jl
+++ b/src/requests/workspace.jl
@@ -109,7 +109,7 @@ function request_julia_config(server::LanguageServerInstance, conn)
         ConfigurationItem(missing, "julia.lint.disabledDirs")
         ]))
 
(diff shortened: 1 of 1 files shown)


ZacLN:
Should this continue to check against 'vscode' as well to continue to support older versions?

davidanthoff:
I'm modestly certain that this string originates from the `vscode-languageclient` v7.0.0 npm package and not from VS Code itself. If that is true, we should be good because we control which version of that package we ship. I think I also bumped the min version of VS Code that we run on as well for the pkg tag stuff, so I think we should be fairly safe that this will never run with older versions.

Having said that, having both in probably also wouldn't do any harm :)

PR #868, opened 2021-01-24
add msg handlers for setTrace

ZacLN:
I can't recall where the setTraceNotification code came from - maybe OK to remove that?


davidanthoff:
> setTraceNotification 

Yeah, it doesn't seem to be part of LSP, right? But lets merge this first to fix things.

davidanthoff (approved):

PR #875, opened 2021-01-30
make oos for getModuleAt non-fatal

pfitzseb:


davidanthoff:
Do we really need this? Can't we just handle the specific error in the client? At the end of the day it _is_ an error condition, and I think that would be cleaner than using a sentinel value here.

PR #877, opened 2021-02-01
Fix get_offset2 calculation

davidanthoff:
This hopefully fixes a bug from crash reporting.

### DIFF
--- a/src/document.jl
+++ b/src/document.jl
@@ -113,23 +113,19 @@ function get_offset2(doc::Document, line::Integer, character::Integer)
     pos = line_offset
 
     while character > 0
-        try
-            if UInt32(text[pos]) >= 0x010000
-                character -= 2
-            else
-                character -= 1
-            end
-        catch err
-            if err isa BoundsError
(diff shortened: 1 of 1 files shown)


src/document.jl
```diff
@@ -113,24 +113,20 @@ function get_offset2(doc::Document, line::Integer, character::Integer)
...
+        if UInt32(text[pos]) >= 0x010000
+            character -= 2
+        else
+            character -= 1
+        end
+
         pos = nextind(text, pos)
     end
 
-    ret = min(pos, next_line_offset)
```
davidanthoff, line 132: Note that this line actually was never effective because we returned `pos` rather than `ret` from the function, which was a bug in itself. The new code with the `if` statement in the loop should be equivalent to this, _and_ also run actually :)

pfitzseb (approved):

src/document.jl
davidanthoff: Is this actually something the linter could catch? Assigning something to a local variable that then gets never used seems like worthy a warning?

ZacLN: Yeah it is, I'll make PR

davidanthoff: @ZacLN is that something StaticLint.jl could do?

PR #880, opened 2021-02-08
remove cst comparison check

ZacLN:
On each textDocument/didChange we currently check whether the updated EXPR matches the fully reparsed EXPR. We've not hit this check yet and it's expensive - I suggest we remove it if it's not hit on the 1.1 release ver.


pfitzseb (approved):

davidanthoff:
I'm putting this on the Backlog for now, because we probably should wait a few days to see whether we see these errors in crash reporting, right? And if something is on "Next Patch" and approved, I might just accidentally merge it :) @ZacLN whenever you think we _should_ merge it, just put it back on the "Next Patch" milestone (or just merge).

PR #887, opened 2021-02-10
special handling for CompletionContext

ZacLN:
Allow for `nothing` in CompletionContext.triggerCharacter


pfitzseb (approved):
We still think this is a LSP violation though, right?

ZacLN:
Yeah

Issue #681, opened 2020-05-07
Return list of DocumentSymbol for textDocument/documentSymbol instead of SymbolInformation


non-Jedi:
As an example, when browsing this repo and sending `textDocument/documentSymbol` for `src/protocol/basic.jl`:

```elisp
client-request (id:5) Thu May  7 17:43:51 2020:
(:jsonrpc "2.0" :id 5 :method "textDocument/documentSymbol" :params
          (:textDocument
           (:uri "file:///home/adam/repos/LanguageServer.jl/src/protocol/basic.jl")))
```

The `PlainText` and `Markdown` fields of [`const MarkupKinds = (PlainText = "plaintext", Markdown = "markdown")`](https://github.com/julia-vscode/LanguageServer.jl/blob/7259bd65140cd6e727b5938c04e4e355e5381762/src/protocol/basic.jl#L96-L97) are returned as `String` kind (15):

```elisp
      (:name "MarkupKinds" :kind 13 :deprecated :json-false :location
             (:uri "file:///home/adam/repos/LanguageServer.jl/src/protocol/basic.jl" :range
                   (:start
                    (:line 97 :character 6)
                    :end
                    (:line 97 :character 17)))
             :containerName nil)
      (:name "PlainText" :kind 15 :deprecated :json-false :location
             (:uri "file:///home/adam/repos/LanguageServer.jl/src/protocol/basic.jl" :range
                   (:start
                    (:line 97 :character 21)
                    :end
                    (:line 97 :character 30)))
             :containerName nil)
      (:name "Markdown" :kind 15 :deprecated :json-false :location
             (:uri "file:///home/adam/repos/LanguageServer.jl/src/protocol/basic.jl" :range
                   (:start
                    (:line 98 :character 21)
                    :end
                    (:line 98 :character 29)))
             :containerName nil)
```

Similarily, the fields of structs are also included e.g. for [`MarkupContent`](https://github.com/julia-vscode/LanguageServer.jl/blob/7259bd65140cd6e727b5938c04e4e355e5381762/src/protocol/basic.jl#L99-L102) as "Variable" kind (13):

```elisp
      (:name "MarkupContent" :kind 23 :deprecated :json-false :location
             (:uri "file:///home/adam/repos/LanguageServer.jl/src/protocol/basic.jl" :range
                   (:start
                    (:line 100 :character 0)
                    :end
                    (:line 103 :character 3)))
             :containerName nil)
[...]

PR #881, opened 2021-02-08
only provide toplevel symbols in tS/dS in vim

ZacLN:
Fixes #681 for now. It's useful to give vscode all the symbols at it automatically nests them within the relevant blocks

For every PR, please check the following:
- [ ] End-user documentation check. If this PR requires end-user documentation in the Julia VS Code extension docs, please add that at https://github.com/julia-vscode/docs.
- [ ] Changelog mention. If this PR should be mentioned in the CHANGELOG for the Julia VS Code extension, please open a PR against https://github.com/julia-vscode/julia-vscode/blob/master/CHANGELOG.md with those changes.

PR #889, opened 2021-02-10
return DocumentSymbol instead of SymbolInformation

pfitzseb:
Alternative to https://github.com/julia-vscode/LanguageServer.jl/pull/881. Fixes https://github.com/julia-vscode/LanguageServer.jl/issues/681.

Haven't tried this with a non-VSCode client yet. Would also be nice to fine tune the `detail` and `selectionRange` fields a bit. It's also a bit odd that we list function parameters here (or at least different from e.g. TS).

### DIFF
--- a/src/requests/features.jl
+++ b/src/requests/features.jl
@@ -23,7 +23,7 @@ function get_definitions(x::SymbolServer.ModuleStore, tls, server, locations)
     end
 end
 
-function get_definitions(x::Union{SymbolServer.FunctionStore,SymbolServer.DataTypeStore}, tls, server, locations) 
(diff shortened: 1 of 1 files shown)


ZacLN:
Does vscode automatically build a hierarchy if we return `DocumentSymbol`s?

pfitzseb:
Well, this *constructs* a hierarchy, so I don't think VSCode has to do anything here.

ZacLN:
:+1: Didn't see the push! to child part

ZacLN (approved):

davidanthoff:
Can you assign to a milestone, so that I know when to merge it? :)

PR #892, opened 2021-02-13
Completions for unexported symbols

ZacLN:
@pfitzseb, as we discussed the other day - is this what you meant?

EDIT - a description
Completions are now provided for symbols that are not exported by modules in the current namespace. These are only triggered when the partial string has length > 3. Two modes are provided where:
* Completion of an unexported variable `var` from some module `m` results in the automatic inclusion of a using statement `using m: var`. This is included at the start of the current root tree and will be attached to existing `using m:...` if possible; or 
* the completion results in a qualified variable (e.g. `m.var`)


pfitzseb:
Almost. What I had in mind would make this work for _all_ symbols, not only unexported ones (which may or may not be a good idea :P). Basically to counteract people using `using Foo` and then not knowing where a symbol comes from.

This feels pretty good though, so I think we could get this in as-is (although it still needs a setting in the UI, right?).

ZacLN:
That feels a bit much for me personally but no reason to not include it as an option. Would we also want a project level action that applies that style retrospectively?

Agree on merging this as is for now

davidanthoff (changes requested):
Needs a review from @pfitzseb and a merge conflict resolution.

pfitzseb (approved):
Alright, let's get this in!

davidanthoff (changes requested):
Tests fail. Do we maybe need a tag for some dependency for this to work?

davidanthoff (approved):

PR #893, opened 2021-02-14
selection ranges

### LINKED ISSUE #838: Add support for SelectionRanges request
Please consider adding support for https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_selectionRange.

In my experience as an IDE user/implementer, this is probably the most important feature to make writing code pleasant, as, combined with multiple cursors, it sort-of turns any editor into a structural editor.

ZacLN:
Fixes #838 

Not clear how to trigger these on the client side so haven't tested

### DIFF
--- a/src/languageserverinstance.jl
+++ b/src/languageserverinstance.jl
@@ -302,6 +302,7 @@ function Base.run(server::LanguageServerInstance)
     msg_dispatcher[workspace_symbol_request_type] = (conn, params) -> workspace_symbol_request(params, server, conn)
     msg_dispatcher[julia_refreshLanguageServer_notification_type] = (conn, params) -> julia_refreshLanguageServer_notification(params, server, conn)
     msg_dispatcher[julia_getDocFromWord_request_type] = (conn, params) -> julia_getDocFromWord_request(params, server, conn)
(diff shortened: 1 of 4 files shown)


pfitzseb:
You can use `Shift-Alt-Right`/`Shift-Alt-Left` to expand/reduce the smart selection (which is what VSCode calls that feature). Doesn't feel too bad with this PR, imho, although there are a few cases where we iterate through too many exprs.

pfitzseb (dismissed):

ZacLN:
OK, let's add this as in then for the next release and then improve based on feedback (unless you have some specific ideas now)

pfitzseb:
Initially I thought it was annoying that it takes two key strokes to get out of a string for example, but that seems to be fairly standard. So no, I don't think I have any specific ideas :)

davidanthoff:
@ZacLN could you resolve the merge conflict?

davidanthoff (changes requested):
Merge conflict needs resolution.

pfitzseb (approved):

davidanthoff (approved):

PR #895, opened 2021-02-15
Add a check for strings with NULL

davidanthoff:
Helps with https://github.com/julia-vscode/crashreporting-issues/issues/8.

### DIFF
--- a/src/document.jl
+++ b/src/document.jl
@@ -37,6 +37,8 @@ function get_text(doc::Document)
 end
 
 function set_text!(doc::Document, text)
+    # TODO Remove this check eventually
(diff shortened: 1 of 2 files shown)


ZacLN (approved):

src/document.jl
```diff
@@ -127,7 +129,7 @@ function get_offset2(doc::Document, line::Integer, character::Integer)
         pos = nextind(text, pos)
     end
 
-    return pos
+return pos
```
pfitzseb, line 132: These whitespace changes are a regression in DocumentFormat.jl, I think. Should figure out what's going on there...

davidanthoff: Argh, I didn't even notice these when I created the PR!

PR #896, opened 2021-02-16
enable symbol server downloads

ZacLN:


src/requests/workspace.jl
```diff
@@ -115,6 +116,7 @@ function request_julia_config(server::LanguageServerInstance, conn)
 
     new_lint_missingrefs = Symbol(something(response[24], :all))
     new_lint_disableddirs = something(response[25], LINT_DIABLED_DIRS)
+    new_symbolserver_download = something(response[26], true)
```
pfitzseb, line 119: Should maybe disable this by default for now?

davidanthoff:
I think we can close this? Because we actually did new PRs that do exactly the same thing, uh, sometimes I should check open PRs rather than just doing a new one ;)

PR #897, opened 2021-02-19
Safe load

ZacLN:
Fixes https://github.com/julia-vscode/crashreporting-issues/issues/8

### DIFF
--- a/src/requests/features.jl
+++ b/src/requests/features.jl
@@ -80,7 +80,7 @@ function textDocument_definition_request(params::TextDocumentPositionParams, ser
         # TODO: move to its own function
         if valof(x) isa String && sizeof(valof(x)) < 256 # AUDIT: OK
             try
-                if isabspath(valof(x)) && !occursin("\0", valof(x)) && safe_isfile(valof(x))
+                if isabspath(valof(x)) && safe_isfile(valof(x))
                     push!(locations, Location(filepath2uri(valof(x)), Range(0, 0, 0, 0)))
                 elseif !isempty(getpath(doc)) && safe_isfile(joinpath(_dirname(getpath(doc)), valof(x)))
                     push!(locations, Location(filepath2uri(joinpath(_dirname(getpath(doc)), valof(x))), Range(0, 0, 0, 0)))
--- a/src/requests/features.jl
+++ b/src/requests/features.jl
@@ -55,7 +55,7 @@ function get_definitions(x::EXPR, tls::StaticLint.Scope, server, locations)
     end
 end
 
-safe_isfile(s) = false
+safe_isfile(s::Symbol) = safe_isfile(string(s))
 function safe_isfile(s::AbstractString)
     try
         !occursin("\0", s) && isfile(s)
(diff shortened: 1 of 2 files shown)


pfitzseb:
I think we should switch to safe `isfile` implementation here that catches errors, no? We can never be sure that `isfile` doesn't throw.

src/requests/features.jl
```diff
@@ -74,9 +80,9 @@ function textDocument_definition_request(params::TextDocumentPositionParams, ser
         # TODO: move to its own function
         if valof(x) isa String && sizeof(valof(x)) < 256 # AUDIT: OK
             try
-                if isabspath(valof(x)) && !occursin("\0", valof(x)) && isfile(valof(x))
+                if isabspath(valof(x)) && !occursin("\0", valof(x)) && safe_isfile(valof(x))
```
pfitzseb, line 83:
```suggestion
                if isabspath(valof(x)) && safe_isfile(valof(x))
```

src/staticlint.jl
```diff
@@ -3,7 +3,7 @@ import StaticLint: getpath, getroot, setroot, getcst, setcst, semantic_pass, get
 hasfile(server::LanguageServerInstance, path::String) = !isempty(path) && hasdocument(server, URI2(filepath2uri(path)))
 function canloadfile(server::LanguageServerInstance, path::String)
     try
-        return !isempty(path) && isfile(path)
+        return !isempty(path) && !occursin("\0", path) && isfile(path)
```
pfitzseb, line 6:
```suggestion
        return !isempty(path) && !safe_isfile(path)
```

src/requests/features.jl
```diff
@@ -59,6 +55,16 @@ function get_definitions(x::EXPR, tls::StaticLint.Scope, server, locations)
     end
 end
 
+safe_isfile(s) = false
+function safe_isfile(s::AbstractString)
+    try
+        !occursin("\0", s) && isfile(s)
+    catch err
+        isa(err, Base.IOError) || isa(err, Base.SystemError) || rethrow()
```
pfitzseb, line 63: Do we really need to rethrow anything here? I don't really see any cases where we can fix anything that causes `isfile` to fail, right?

src/requests/features.jl
```diff
@@ -59,6 +55,16 @@ function get_definitions(x::EXPR, tls::StaticLint.Scope, server, locations)
     end
 end
 
+safe_isfile(s) = false
```
davidanthoff, line 58: What is the purpose of this method? I think it would be safer to see a method error if someone tries to call this function with something other than an `AbstractString`, because that always seems like a bug to me?

src/requests/features.jl
davidanthoff: I think we want to see what error conditions might occur in crash reporting, so I generally think we should only "mask" exceptions that we understand and then rethrow anything else so that we see it in crash reporting and can then make a decision whether we also want to mask it, or whether there is some other root cause for it.

src/requests/features.jl
davidanthoff: @ZacLN could you take a look at this method here? I think once we have sorted this out, we could merge this?

pfitzseb: Removed this and added a `safe_isfile(s::Symbol)` method, because that one actually kinda makes sense.

davidanthoff (approved):

ZacLN:
Thanks @pfitzseb !

PR #898, opened 2021-02-19
Add a diagnostic for relative paths

davidanthoff:
Helps diagnosic https://github.com/julia-vscode/crashreporting-issues/issues/26.

We somehow end up with a relative path for a document there, so we need to figure out how that is slipping in.

- [X] End-user documentation check. If this PR requires end-user documentation in the Julia VS Code extension docs, please add that at https://github.com/julia-vscode/docs.
- [X] Changelog mention. If this PR should be mentioned in the CHANGELOG for the Julia VS Code extension, please open a PR against https://github.com/julia-vscode/julia-vscode/blob/master/CHANGELOG.md with those changes.

### DIFF
--- a/src/document.jl
+++ b/src/document.jl
@@ -13,7 +13,7 @@ mutable struct Document
     root::Document
     function Document(uri::AbstractString, text::AbstractString, workspace_file::Bool, server=nothing)
         path = something(uri2filepath(uri), "")
-        isabspath(path) || throw(LSRelativePath("Relative path `$path` is not valid."))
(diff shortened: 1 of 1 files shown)


pfitzseb:
I don't think this is going to help too much, but can't hurt to get this in.

pfitzseb (approved):

davidanthoff:
I think it should be good enough to figure out whether we are getting the faulty path from the client or StaticLint.jl, right?

PR #900, opened 2021-02-19
don't attach user code to error

ZacLN:
Fixes #.


src/requests/hover.jl
```diff
@@ -46,7 +46,7 @@ function get_hover(b::StaticLint.Binding, documentation::String, server)
                 end
             end
         else
-            try
+            documentation = try
```
davidanthoff, line 49: Is that needed? Aren't all code paths already assigning things to `documentation`?

davidanthoff (approved):

PR #905, opened 2021-03-03
better getModuleAt

pfitzseb:
by ensuring that the offset is inside (not on the edge of) the EXPR's span.

Fixes https://github.com/julia-vscode/julia-vscode/issues/1600 and similar issues (at least when the cursor actually is *outside* of the module definition).

### DIFF
--- a/src/requests/features.jl
+++ b/src/requests/features.jl
@@ -272,8 +272,8 @@ function julia_getModuleAt_request(params::VersionedTextDocumentPositionParams, 
     if hasdocument(server, uri)
         doc = getdocument(server, uri)
         if doc._version == params.version
-            offset = get_offset2(doc, params.position.line, params.position.character, true)
-            x = get_expr(getcst(doc), offset)
+            offset = get_offset2(doc, params.position.line, params.position.character)
+            x = get_expr_or_parent(getcst(doc), offset - 1)
             if x isa EXPR
                 scope = StaticLint.retrieve_scope(x)
                 if scope !== nothing
--- a/src/requests/features.jl
+++ b/src/requests/features.jl
@@ -272,7 +272,7 @@ function julia_getModuleAt_request(params::VersionedTextDocumentPositionParams, 
     if hasdocument(server, uri)
         doc = getdocument(server, uri)
         if doc._version == params.version
-            offset = get_offset2(doc, params.position.line, params.position.character)
+            offset = get_offset2(doc, params.position.line, params.position.character, true)
             x = get_expr_or_parent(getcst(doc), offset - 1)
             if x isa EXPR
                 scope = StaticLint.retrieve_scope(x)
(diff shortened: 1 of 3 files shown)


src/requests/features.jl
```diff
@@ -272,8 +272,8 @@ function julia_getModuleAt_request(params::VersionedTextDocumentPositionParams,
     if hasdocument(server, uri)
         doc = getdocument(server, uri)
         if doc._version == params.version
-            offset = get_offset2(doc, params.position.line, params.position.character, true)
-            x = get_expr(getcst(doc), offset)
+            offset = get_offset2(doc, params.position.line, params.position.character)
+            x = get_expr_or_parent(getcst(doc), offset - 1)
```
pfitzseb, line 276: I'm pretty sure there's an off-by-one error somewhere between the CST and what `get_offset` returns, which might very well cause issues in other cases...

davidanthoff: Do you have a repo case? I think it would be super valuable for us to fix these kind of problems for good...

davidanthoff (approved):

src/requests/features.jl
pfitzseb: It's a systematic error/difference. When you have the cursor at position `(0,0)` then `get_offset` returns `1`, but we usually start indexing the position in the CST from `0`. Neither is incorrect, but afaict we're inconsistent about it.

davidanthoff: So `get_offset2` is all around 1-based, I think. And then the idea is that it designates the position to the _left_ of the indexed byte. So generally, the lowest offset that we ever need/get is 1, because that designates the point to the left of the first character in any string. But we _can_ get offset positions that are larger than the length of a string, because when we designate a range say for the entire content of a string, then the stop bye position will be `lastbyte+1`.

I think that all follows the conventions on the VS Code side of things and the LSP, with the exception that our offsets internally are 1 rather than 0 based.

I have to admit I don't really understand why CST would ever return an offset of `0` :) If that is a 1-based offset, then that would amount to a `-1` offset in 0-based land, which I don't understand why we would ever want that?

pfitzseb: No, I think we just usually iterate over the text with a zero-based offset (because that's much more natural for an *offset* as opposed to an index). See e.g. the definition of `get_expr`, which includes a `pos=0` default argument. I've added a comment and made the code a bit clearer though.

davidanthoff: So I think that means that `get_offset2` is probably misnamed, and it should really be named `get_string_index` or something like that, because it most definitely returns an index and not an offset in the way you defined it (which makes sense to me).

pfitzseb: Yup. Let's tackle that in a follow up though.

davidanthoff (approved):

PR #906, opened 2021-03-04
allow fuzzy completions

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/issues/1988. 
image

Do we want to add an opt-out setting to this? Imho this is unintrusive enough not to need one (JS completions are always fuzzy as well). 


ZacLN:
I agree - reckon it's Ok without and can reconsider if we get complaints

ZacLN (approved):

PR #907, opened 2021-03-10
better shutdown handling

pfitzseb:
~Needs https://github.com/julia-vscode/JSONRPC.jl/pull/44.~

### DIFF
--- a/src/languageserverinstance.jl
+++ b/src/languageserverinstance.jl
@@ -259,7 +259,7 @@ function Base.run(server::LanguageServerInstance)
     trigger_symbolstore_reload(server)
 
     @async try
-        while true
+        while isopen(server.jr_endpoint)
             msg = JSONRPC.get_next_message(server.jr_endpoint)
             put!(server.combined_msg_queue, (type = :clientmsg, msg = msg))
         end
@@ -270,10 +270,13 @@ function Base.run(server::LanguageServerInstance)
         else
             Base.display_error(stderr, err, bt)
         end
+    finally
+        put!(server.combined_msg_queue, (type = :close,))
+        close(server.combined_msg_queue)
     end
 
     @async try
-        while true
+        while isopen(server.symbol_results_channel)
             msg = take!(server.symbol_results_channel)
             put!(server.combined_msg_queue, (type = :symservmsg, msg = msg))
         end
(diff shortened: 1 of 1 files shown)


src/languageserverinstance.jl
```diff
@@ -259,7 +259,7 @@ function Base.run(server::LanguageServerInstance)
     trigger_symbolstore_reload(server)
 
     @async try
-        while true
+        while isopen(server.jr_endpoint)
             msg = JSONRPC.get_next_message(server.jr_endpoint)
             put!(server.combined_msg_queue, (type = :clientmsg, msg = msg))
         end
```
davidanthoff, line 265: This strikes me as a race condition. Don't we have to make sure that `get_next_message` also doesn't get stuck if the endpoint is closed? And if we do that, we might as well not have the `isopen` call in the `while` loop because that situation would already be handled by the `get_next_message`?

src/languageserverinstance.jl
```diff
@@ -270,10 +270,13 @@ function Base.run(server::LanguageServerInstance)
...
+    finally
+        put!(server.combined_msg_queue, (type = :close,))
+        close(server.combined_msg_queue)
     end
 
     @async try
-        while true
+        while isopen(server.symbol_results_channel)
             msg = take!(server.symbol_results_channel)
```
davidanthoff, line 280: Same here, I think in general these kind of checks whether something is in the right state or not don't work well in async/parallel code because we should never really assume that things don't change between that check and the next thing we call. So we need to handle that situation in the next line (280) in any case, so we might as well _only_ handle it there.

davidanthoff (changes requested):
So I'm not really convinced this is the right way to handle this, see my comments inline.

src/languageserverinstance.jl
pfitzseb: Right, these two `isopen` checks are basically unnecessary. `take!` will always either return or throw an exception if the underlying channel gets closed while `wait`ing for something to happen, so the important logic here is in the `finally` block, which closes the aggregate channel.

pfitzseb:
This should be fairly uncontroversial as-is, right?

Would like to get this in asap because I've started collecting zombies too and testing this requires a bit of time, I think
image

davidanthoff (approved):

PR #908, opened 2021-03-11
Add support for textDocument/documentHighlight request

jwortmann:

***

This allows to highlight other occurences of the identifier at a given cursor position, see [Document Highlights Request](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_documentHighlight).
The implementation is quite simple and similar to the textDocument/references request. Not sure whether `StaticLint.hasbinding` is the best way to distinguish between write and read access, but it seems to work well for me so far (okay, it doesn't recognize e.g. `push!` as write access, but at least is a good enough heuristics).

Here is an illustration from Sublime Text, with a green background for read access, and orange background for write access.

highlight

I don't know how to test this with VS Code, but it should most likely work without further ado. I think the setting "editor.occurrencesHighlight" controls whether these highlights are enabled or not.

It doesn't work entirely correct whenever a variable gets assigned to multiple times, but I can see the same behavior for textDocument/references (both use the same logic from `StaticLint.refof`, so that seems to be the cause of this issue and should probably be fixed there, if it is a bug):

problem

I also saw that there is an outdated draft PR #510 for this LSP request, but I believe the implementation therein does not what the request is actually meant to be used for, according to the specs (and how most other language servers use it).


pfitzseb (commented):
Nice, this works fine in VSCode as well.

I'd prefer less code duplication between `find_references` and `textDocument_documentHighlight_request`, so a bit of refactoring would be good. Can do that after merging this though.

pfitzseb (approved):

pfitzseb:
Should merge https://github.com/julia-vscode/LanguageServer.jl/pull/911 after.

PR #935, opened 2021-05-07
fix 280

ZacLN:
Fixes https://github.com/julia-vscode/StaticLint.jl/issues/280

Previously we were pushing arguments to the top expresions .args field directly, rather than using the specialised `push!(expr::EXPR,item)`. The specialised version makes sure the parent of each `item` is set to `expr`.


pfitzseb (approved):
Yep, this fixes the issue.

PR #939, opened 2021-05-18
Report file when parse error occurs

IanButterworth:
Reports the file when a parsing error occurs, to aid in tracking down the issue


pfitzseb (approved):
Can't hurt to have this, although I'm not 100% sure it'll always be correct. Also won't show up in our crash reports, but that seems fine considering we'd need the user for the file anyways.

PR #940, opened 2021-05-18
Multithreaded load_folder

IanButterworth:
While investigating the bug in #939 I wondered if multithreading `load_folder` would work.

It seems to.. but needs further testing

This implements per-file multithreading in `load_folder`, but also (inadvertently) turns parsing errors into soft-errors, so that the language server doesn't crash when hitting a problematic file.

It still prints the file name and full stacktrace, so could replace #939

### DIFF
--- a/src/requests/init.jl
+++ b/src/requests/init.jl
@@ -89,6 +89,7 @@ end
 function load_folder(path::String, server)
     if load_rootpath(path)
         try
+            sema = Base.Semaphore(Threads.nthreads())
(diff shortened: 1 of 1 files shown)


ZacLN:
`StaticLint.semantic_pass` (called by `parse_all`) is most definitely not thread safe, as it shares a graph of data between files (e.g. scopes, bindings/references). It could feasibly be done at the workspace folder level as you suggest but I'm not sure what would be gained (i.e. likely no speedup as most of initial run time is code building/compilation)

PR #963, opened 2021-07-14
add rd/prepareRename

ZacLN:
Fixes https://github.com/julia-vscode/julia-vscode/issues/2249


src/requests/init.jl
```diff
@@ -21,7 +21,7 @@ const serverCapabilities = ServerCapabilities(
     true,
     false,
     missing,
-    true,
+    RenameOptions(missing, true),
```
pfitzseb, line 24: This needs to be explicitly supported by the client:
> RenameOptions may only be specified if the client states that it supports prepareSupport in its initial initialize request.

pfitzseb (approved):

davidanthoff:
Could we add a test that tests the actual implementation?

PR #964, opened 2021-07-16
Normalize project path in LanguageServerInstance constructor

fredrikekre:


fredrikekre:
Actually, the canonical way of specifying an env is the path to the toml file, so I am happy with just https://github.com/julia-vscode/SymbolServer.jl/pull/227

PR #967, opened 2021-07-20
remove hover message for closing function call parenthesis

pfitzseb:
because these are already indicated by highlighting the matching paren anyways.

### DIFF
--- a/src/requests/hover.jl
+++ b/src/requests/hover.jl
@@ -221,10 +221,6 @@ function get_closer_hover(x::EXPR, documentation)
             else
                 documentation = "Closes `$(headof(parentof(x)))` expression."
             end
-        elseif headof(x) === :RPAREN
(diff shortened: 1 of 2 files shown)


ZacLN:
Happy for this to be removed but I think worth pointing out that I added this for the case where the opening paren isn't on the same page

davidanthoff (changes requested):
Tests fail, and @pfitzseb do you think this should be merged in light of @ZacLN's comment?

pfitzseb:
Tests should pass now. And yes, in my opinion we should remove this; `Go To Bracket` almost provides the same functionality (imho it's more useful because you can also tell *which* `for` loop or whatever the current end closes). In addition, this makes inline results much nicer.

davidanthoff (approved):

PR #972, opened 2021-08-02
Use JuliaFormatter instead of DocumentFormat

pfitzseb:
Fixes https://github.com/julia-vscode/DocumentFormat.jl/issues/70.


pfitzseb:
Requires https://github.com/domluna/JuliaFormatter.jl/pull/448 for compat with older Julia versions.

pfitzseb:
This now also adds some rudimentary support for selection formatting.

davidanthoff (changes requested):
So I think the main blocker at the moment is that we're waiting for `MinimalStyle`, i.e. https://github.com/domluna/JuliaFormatter.jl/pull/256, right?

davidanthoff:
Oh, and also, it would be great if @ZacLN could generally take a look at JuliaFormatter.jl etc and chime in.

src/requests/features.jl
```diff
@@ -114,15 +114,116 @@ function get_file_loc(x::EXPR, offset=0, c=nothing)
...
+                idx += 1
+            else
+                break
+            end
+        end
+        line = line[1:idx]
+        longest_prefix = CSTParser.longest_common_prefix(something(longest_prefix, line), line)
+    end
+
+    config = get_juliaformatter_config(doc, server)
```
ZacLN, line 201: Does this request get hit every key stroke? If so it's not ideal to search the disk each time. I can't remember whether we're able to have a file watch from within the languageserver which may be why this has been done

ZacLN (changes requested):

src/requests/features.jl
davidanthoff: We could just add those files to https://github.com/julia-vscode/julia-vscode/blob/master/src/extension.ts#L232, and then we would get normal LSP messages about their content etc. That seems in general better?

pfitzseb: > Does this request get hit every key stroke?

No, only when explicitly requested.

> We could just add those files to https://github.com/julia-vscode/julia-vscode/blob/master/src/extension.ts#L232, and then we would get normal LSP messages about their content etc. That seems in general better?

IIUC we'd only be notified by FS events or when the user opens/changes the `.JuliaFormatter.toml`. The client won't just dump all matching files into the server.

PR #977, opened 2021-08-05
Harden against files with embedded NULL

davidanthoff:
Fixes https://github.com/julia-vscode/crashreporting-issues/issues/46.

### DIFF
--- a/src/requests/workspace.jl
+++ b/src/requests/workspace.jl
@@ -15,7 +15,7 @@ function workspace_didChangeWatchedFiles_notification(params::DidChangeWatchedFi
                     filepath = uri2filepath(uri)
                     content = try
                         s = read(filepath, String)
-                        if !isvalid(s)
+                        if !isvalid(s) || occursin('\0', s)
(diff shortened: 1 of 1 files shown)


pfitzseb (approved):
Still not sure how this can happen.

davidanthoff:
Just some random piece of code writing an invalid file to disc, right?

pfitzseb:
Hm, fair. I thought VSCode would handle that case, but I guess JS strings can contain `NULL`s just fine.

davidanthoff:
This is a load path where we load the file content directly from disc, VSC is only telling us that the file changed on disc, but completely out of the picture when it comes to the file content. That whole design is pretty broken because VS Code now supports workspaces that don't exist on disc at all, and I think they are working on some LSP version where we essentially don't have to use file IO functions in the LS at all but can get all content from the client via the protocol, but not entirely sure what the status on that is.

pfitzseb:
Oh, I could've just read the `read(file, String)` line :P Didn't actually realize this happened anywhere...

PR #980, opened 2021-08-09
Autocomplete of methods for given variables

### LINKED ISSUE #979: Autocomplete of methods for given variables
Recently, the PR https://github.com/JuliaLang/julia/pull/38791 for the REPL got merged to be included in the next Julia version. This means that the REPL will tab-complete statements like `?(x, y)` to yield all methods that apply to `x` and `y` as well as `?(x` with arbitrary additional arguments. It would be nice to have this also in the language server for use in VSCode and other editors. The current discussion is on https://discourse.julialang.org/t/tab-completion-to-find-methods-for-given-variable-s/66076

krystophny:
Fixes #979.


krystophny:
The initial version is working. Up to now, it has autocomplete for functions available from the SymbolServer that have methods that exactly match the type of a single first argument. Next steps should be to support

- [ ] User-defined functions in outside SymbolServer database (current scope / Main)
- [ ] Matching functions with multiple arguments

@pfitzseb could you have a look if the general idea and functionality is OK? I won't have much time in the coming days to continue, but maybe this may serve as a basis for further work. Do you think one could merge also such a "minimal version" in case the other features are much harder to realize?

src/requests/completions.jl
```diff
@@ -540,3 +548,62 @@ function get_tls_arglist(tls::StaticLint.Scope)
         error()
     end
 end
+
+function method_completion(x, state, xlen)
+    scope = scopeof(parentof(parentof(state.x)))
```
ZacLN, line 553: I think we need to check `scope isa Scope`

krystophny:
Closing, since not followed up for more than 3 years.

PR #990, opened 2021-11-04
Fix document symbols for function-like objects

BenPH:
Fixes https://github.com/julia-vscode/julia-vscode/issues/2303

Function-like objects were being rejected as valid bindings by `is_valid_binding_name` and so all symbols in their scope were escaping.

I don't know if it's the most intuitive but right now I have on `::TheType` for the name given to callable object symbols. 

image

### DIFF
--- a/src/requests/features.jl
+++ b/src/requests/features.jl
@@ -4,7 +4,7 @@
 resolve_shadow_binding(b) = b
 function resolve_shadow_binding(b::StaticLint.Binding, visited=StaticLint.Binding[])
     if b in visited
-        throw(LSInfiniteLoop("Inifinite loop in bindings."))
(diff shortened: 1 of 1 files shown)


pfitzseb:
Cool, thanks for the fix! Would you mind adding a test [hereabouts](https://github.com/julia-vscode/LanguageServer.jl/blob/4fca88cecc82fb7c6520881b10f5c3894ccf8506/test/requests/features.jl#L93-L100) for this?

pfitzseb (approved):

pfitzseb:
Added a test in https://github.com/julia-vscode/LanguageServer.jl/commit/915c4627d780f2043be28fa40b3bff473dd1cc2e.

PR #1015, opened 2021-12-31
Fix hover docs for definitions in nested macros

BenPH:
Fixes https://github.com/julia-vscode/julia-vscode/issues/2650

This assumes that if a docstring is present, then no matter what the preceding macros do, the docstring should still apply to the defined struct/method/binding.

```julia
"docstring"
@macro1 @macro2 foo(x) = x + 1 # assume "docstring" should still apply to foo(x)
```

Also adds a missing newline for hovers with single line docstrings on bindings. e.g
```julia
"docstring"
x = 123
```

### DIFF
--- a/src/requests/hover.jl
+++ b/src/requests/hover.jl
@@ -55,7 +55,7 @@ function get_hover(b::StaticLint.Binding, documentation::String, server)
                 else
                     documentation
                 end
-                documentation = string(documentation, "```julia\n", prettify_expr(to_codeobject(b.val)), "\n```\n")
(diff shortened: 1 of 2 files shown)


pfitzseb (approved):
https://github.com/julia-vscode/LanguageServer.jl/pull/1015/files#diff-582dc9e84064c9ee934a21d75734c7d15d4d2aec8045f791f3837cc7b235eab9R150 doesn't make me super happy, but I think this is totally fine :)

Thanks!

BenPH:
Yeah.. I assume it's rare to have more than a few nested macros so hopefully it shouldn't matter too much.

PR #1028, opened 2022-02-28
Remove file format options to avoid JuliaFormatter crash

### LINKED ISSUE #1027: formatting fails with verbose, overwrite or format_markdown in config file
Formatting a file fails if verbose, format_markdown or overwrite options are at  JuliaFormatter.toml 

```
[Error - 10:27:11 AM] Request textDocument/formatting failed.
  Message: Failed to format document: MethodError(Core.var"#Type##kw"(), ((always_use_return = true, short_to_long_function_def = false, whitespace_ops_in_indices = false, remove_extra_newlines = true, always_for_in = true, import_to_using = true, pipe_to_function_call = true, whitespace_in_kwargs = false, annotate_untyped_fields_with_any = false, conditional_to_if = true, indent_submodule = true, whitespace_typedefs = true, indent = 4, align_struct_field = true, align_assignment = true, align_conditional = true, format_docstrings = false, verbose = true, align_pair_arrow = true, format_markdown = true, margin = 94), JuliaFormatter.Options), 0xffffffffffffffff).
  Code: -32000 
```

dpinol:
verbose, format_markdown & overwrite options are not filtered out from JuliaFormatter.toml because they only make sense for JuliaFormat functions to format a file not for formatting a text. Passing them makes format_text crash

Fixes #1027.

### DIFF
--- a/src/requests/features.jl
+++ b/src/requests/features.jl
@@ -156,11 +156,7 @@ function textDocument_formatting_request(params::DocumentFormattingParams, serve
     config = get_juliaformatter_config(doc, server)
 
     newcontent = try
-        if config === nothing
-            JuliaFormatter.format_text(get_text(doc); default_juliaformatter_config(params)...)
-        else
-            JuliaFormatter.format_text(get_text(doc); JuliaFormatter.kwargs(config)...)
-        end
+        format_text(get_text(doc), params, config)
     catch err
         return JSONRPC.JSONRPCError(
             -32000,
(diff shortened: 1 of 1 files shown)


domluna:
this looks right to me, since the context of the formatter with vscode is on individual files.

I guess one thing to consider is if the file is .md or .jmd to allow formatting with format_markdown but that could be a separate issue since it probably requires more thought

domluna:
the JuliaFormatter docs are outdated atm, not sure what's going on there and I keep forgetting to investigate

dpinol:
> Alright. Docs could do with adjusting then :)
> 
> For this PR: Let's filter the allowed names by `fieldnames(JuliaFormatter.Options)`?

@pfitzseb  done (amending the previous commit)

pfitzseb (approved):
Thanks for the patch!

Issue #750, opened 2020-06-13
SymbolServer must be imported into Main before LanguageServerInstance can be successfully run


ExpandingMan:
I'm working on [making sure LanguageServer always gets initialized properly in the appropriate environment](https://github.com/ExpandingMan/LSPNeovim.jl).

I'm often discovering the following behavior: the `LanguageServerInstance` seems to start fine.  After a few minutes, when I'd normally see the "receiving language from SymbolServer" message, I instead get the following (prior messages included here also, apologies for the terrible formatting)
```
"[ Info: Activating LSPNeovim environment\n"
" Activating environment at `~/.cache/nvim/nvim_lsp/LSPNeovim/Project.toml`\n"
'┌ Info: Initializing Language Server\n│   pwd() = "/home/expandingman/.julia/dev/LSPNeovim/bin"\n│   env = "/home/expandingman/.julia/dev/LSPNeovim/bin/.."\n└   depot = "/home/expandingman/.julia"\n'
"ERROR: "
"UndefVarError: SymbolServer not defined"
"\nStacktrace:"
"\n"
" [1] "
"deserialize_module"
"("
"::"
"Serialization"
".Serializer{IOStream}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Serialization/src/Serialization.jl:956\n [2] handle_deserialize(::Serialization.Serializer{IOStream}, ::Int32) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Serialization/src/Serialization.jl:855\n [3] deserialize(::Serialization.Serializer{IOStream}"
") at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Serialization/src/Serialization.jl:773\n [4] deserialize_datatype(::Serialization.Serializer{IOStream}, ::Bool"
") at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Serialization/src/Serialization.jl:1251\n [5] handle_deserialize(::Serialization.Serializer{"
"IOStream}, ::Int32) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Serialization/src/Serialization.jl:826\n [6] deserialize(::"
"Serialization.Serializer{IOStream}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Serialization/src/Serialization.jl:773"
[...]

PR #1036, opened 2022-03-14
remove using SymbolServer from docs

ChrHorn:
Seems this is no longer needed.

https://github.com/julia-vscode/LanguageServer.jl/issues/750


pfitzseb (approved):
Thanks!

PR #1038, opened 2022-03-19
Fix a LSP protocoll definition

davidanthoff:
Fixes a crash report. `null` is a valid return type according to the spec.


davidanthoff:
I'm merging, want to tag a new patch release and this seems harmless enough.

PR #1044, opened 2022-03-29
Large file limit for linting

pfitzseb:
and better startup diagnostics. Requires https://github.com/julia-vscode/StaticLint.jl/pull/333.


davidanthoff (approved):
I think we are more or less setup to also send performance measurements inside the LS via telemetry. Might require a little bit of extra work, but that would allow us to get a sense of what is generally going on with these things in the user base.

PR #1046, opened 2022-04-03
Completion for string macros

fredrikekre:
This implements completions for string macros to allow them to complete
without the initial `at` and trailing `_str`. For example, `artif`
completes to `artifact"` like in the REPL. If the completion mode is set
to `:import` the real name is imported (`using Pkg: at-artifact_str`),
and when the completion mode is `:qualify` the qualified string macro is
inserted (`Pkg.artifact"`).

### DIFF
--- a/src/requests/completions.jl
+++ b/src/requests/completions.jl
@@ -164,32 +164,60 @@ function texteditfor(state::CompletionState, partial, n)
     TextEdit(Range(Position(state.range.start.line, max(state.range.start.character - length(partial), 0)), state.range.stop), n)
 end
 
+function string_macro_altname(s)
+    if startswith(s, "@") && endswith(s, "_str")
+        return chop(s; head=1, tail=4) * '"'
+    else
+        return nothing
+    end
+end
+
 function collect_completions(m::SymbolServer.ModuleStore, spartial, state::CompletionState, inclexported=false, dotcomps=false)
+    possible_names = String[]
     for val in m.vals
         n, v = String(val[1]), val[2]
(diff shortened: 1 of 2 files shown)


src/requests/completions.jl
```diff
@@ -168,12 +168,16 @@ function collect_completions(m::SymbolServer.ModuleStore, spartial, state::Compl
     for val in m.vals
         n, v = String(val[1]), val[2]
         (startswith(n, ".") || startswith(n, "#")) && continue
+        n_for_import = n
+        if startswith(n, "@") && endswith(n, "_str")
+            n = chop(n; head=1, tail=4) * '"'
```
fredrikekre, line 173: I think this breaks completion of `@xxx_str`, will fix.

fredrikekre: Hmm okay, so there is also a fuzzy score to check matching, not just exact matches, so I guess one would have to push two possible completions for this case.

fredrikekre: I now keep a vector of possible insertions per "real" name, and push them all to the candidates. Perhaps a bit convoluted for such a small feature but..

pfitzseb:
[`collect_completions(x::StaticLint.Scope, ...`](https://github.com/julia-vscode/LanguageServer.jl/blob/c2703f407af9cac637df598146e23ea73a530a61/src/requests/completions.jl#L249-L262) needs the same treatment, otherwise
```
macro fooo_str(ex) ex end
foo<tab>
```
won't work.

fredrikekre:
Done

pfitzseb (approved):

PR #1048, opened 2022-04-05
Implement code action to replace ==/!= with ===/!==

fredrikekre:
This implements a new code action, CompareNothingWithTripleEqual, which
offers to replace == and != with === and !==, respectively, for the
corresponding linter warning(s).

### DIFF
--- a/src/requests/actions.jl
+++ b/src/requests/actions.jl
@@ -267,6 +267,15 @@ function remove_farg_name(x, server, conn)
     JSONRPC.send(conn, workspace_applyEdit_request_type, ApplyWorkspaceEditParams(missing, WorkspaceEdit(missing, TextDocumentEdit[tde])))
 end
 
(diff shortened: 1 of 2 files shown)


fredrikekre:
From https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeAction it looks like the response can also contain the diagnostics that the edit resolves? I am not sure how to pass that, and if it is necessary (the file will be relinted anyway, is it just so users can see that it actually resolves something?

fredrikekre:
I played around a bit more but I think that linking diagnostics to fixes like that require quite some more changes (e.g. returning `CodeAction`s instead of `Command`s from `codeAction` request, and support for `data` in various structs, which is in version 3.16) so I think that can be done in a future PR.

pfitzseb (approved):

PR #1050, opened 2022-04-08
Fix location lookup for code actions, fixes #1047.

### LINKED ISSUE #1047: Locations for code actions are "off-by-one"
Two examples:
With the cursor on `g`, the `Expand function definition` action expands `f` instead of `g`
```julia
f(x) = x
g(x) = x
```

With the cursor on `u`, the `Delete name of unused argument` action doesn't show up, you have to go to `n` (this is particularly annoying in neovim, since if you jump between diagnostics you end up at the first character):
```julia
function f(unused)
    return nothing
end
```

fredrikekre:


src/requests/actions.jl
```diff
@@ -7,7 +7,7 @@ end
 function textDocument_codeAction_request(params::CodeActionParams, server::LanguageServerInstance, conn)
     commands = Command[]
     doc = getdocument(server, URI2(params.textDocument.uri))
-    offset = get_offset(doc, params.range.start) # Should usef get_offset2?
+    offset = get_offset2(doc, params.range.start)
     x = get_expr(getcst(doc), offset)
     arguments = Any[params.textDocument.uri, offset] # use the same arguments for all commands
```
davidanthoff, line 12: So, I never understood what exactly `get_offset` returns, but I always thought that it was _the thing_ that things like `get_expr` expect...

`get_offset2` returns a valid index into the underlying string that holds the document, and I primarily introduced it to fix the text sync implementation. But I'm not sure we can just swap out `getoffset2` for `getoffset` in situations where we then pass the return value onto functionality that was built assuming the `get_offset` semantics?

But take this just as a heads-up about the history, this fix might also just be correct, I don't really know anything about `get_expr` or how it works :)

fredrikekre: Honestly I didn't look into the details, but while debugging I noticed the comment and just tried it out. It seems to fix all the issues I had at least.

PR #1052, opened 2022-04-08
Show type in completion details

pfitzseb:
and hover.

image
image

ToDo:
- [x] Tests
- [ ] improve `completion_kind` heuristics


fredrikekre:
Tested your example in neovim:

Would be nice to get both fields to show the same kind?
Screenshot from 2022-04-09 00-08-47

The details look very strange:
Screenshot from 2022-04-09 00-09-12
Screenshot from 2022-04-09 00-09-24

Hover looks like in your example:
Screenshot from 2022-04-09 00-09-43

But completion details again look strange:
Screenshot from 2022-04-09 00-10-27

pfitzseb:
You wanna give this another go, @fredrikekre ? I didn't fix anything related to CompletionKinds yet (well, except for removing the Enum fallback; anything else requires a deep dive into SymbolServer and I don't have the time for that atm), but completions should already be more helpful.

fredrikekre:
Looks good, the strange double type things are gone (`x::Int::Int64` etc).

PR #1063, opened 2022-04-28
Implement basic support for CodeAction.isPreferred

fredrikekre:
This sends the .isPreferred property to the client if it supports is.
Currently only the action replacing == with === is marked as preferred,
since the other current ones are more situational.

### DIFF
--- a/src/requests/actions.jl
+++ b/src/requests/actions.jl
@@ -2,6 +2,7 @@ struct ServerAction
     id::String
     desc::String
     kind::Union{CodeActionKind,Missing}
+    preferred::Union{Bool,Missing}
     when::Function
     handler::Function
 end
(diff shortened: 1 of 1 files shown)


src/requests/actions.jl
```diff
@@ -43,8 +55,12 @@ function textDocument_codeAction_request(params::CodeActionParams, server::Langu
         for (_, sa) in LSActions
             if sa.when(x, params)
                 action = CodeAction(
-                    sa.desc, sa.kind, missing, missing, missing,
-                    Command(sa.desc, sa.id, arguments),
+                    sa.desc, # title
+                    sa.kind, # kind
+                    missing, # diagnostics
+                    client_preferred_support(server) ? sa.preferred : missing, # isPreferred
```
fredrikekre, line 61: Perhaps it is too "strict" to only send this if the client claims to support it -- servers and clients should ignore unknown fields anyway IIUC.

PR #1067, opened 2022-04-29
Use snippet kind for keyword completions which include tab stops

jwortmann:
This PR proposes to use the "Snippet" kind for keyword completions, which include tab stops (`$0`, `$1`, ....) and in most cases multiple lines.

The kind is used for the icon in the completions popup and could for example also be used by clients to filter out snippet completions if users don't want to see them.


pfitzseb (approved):

pfitzseb:
Thanks!

PR #1068, opened 2022-04-29
Run PR-CI on select Julia versions only

pfitzseb:


davidanthoff:
This won't work long term, package butler will immediately try to undo this :) If we want this, we need to update the template in package butler. We can obviously leave for now, until package butler supports something like this.

PR #1071, opened 2022-05-01
New server option to replace unused function argument with underscore

fredrikekre:
This implements a new option, `julia.unusedargument`, which can be set
to either `"type"` (default) or `"underscore"`. This option affects the
replacement offered by the remove-unused-argumentname code action. The
default (`"type"`) is to replace `foo` with `::Any`, and `bar::Int` with
`::Int`. When the option is set to `"underscore"` the offered
replacement is instead `_` and `_::Int`, respectively.

### DIFF
--- a/src/languageserverinstance.jl
+++ b/src/languageserverinstance.jl
@@ -43,6 +43,7 @@ mutable struct LanguageServerInstance
     lint_missingrefs::Symbol
     lint_disableddirs::Vector{String}
     completion_mode::Symbol
+    unused_argument::Symbol
 
     combined_msg_queue::Channel{Any}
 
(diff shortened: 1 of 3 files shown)


pfitzseb:
Hm, how about two distinct actions for this? I personally would like `a::Any` -> `_` and `b::Foo` -> `::Foo`... :)

fredrikekre:
I would also like that actually :D First I thought about implementing some type of template as the option but I don't know exactly what that would look like. Could have two different suggested changes, but would be nice to have a default instead of having to choose every time.

pfitzseb:
Well, we also have the option of just going with what both of us like (without any configuration) :D

fredrikekre:
Sure, lets try that (#1072) and we can implement an option if the people complain.

PR #1074, opened 2022-05-05
Fix missing URI conversion in code action execution.

fredrikekre:
I am confused how CI passed without this, this path should be tested by https://github.com/julia-vscode/LanguageServer.jl/blob/0955a4d6e6bf1f88809054f4b18e9674445f1ad3/test/requests/actions.jl#L8 etc.

### DIFF
--- a/src/requests/actions.jl
+++ b/src/requests/actions.jl
@@ -75,7 +75,7 @@ function textDocument_codeAction_request(params::CodeActionParams, server::Langu
 end
 
 function workspace_executeCommand_request(params::ExecuteCommandParams, server::LanguageServerInstance, conn)
-    uri = params.arguments[1]
+    uri = URI(params.arguments[1])
     offset = params.arguments[2]
     doc = getdocument(server, uri)
     x = get_expr(getcst(doc), offset)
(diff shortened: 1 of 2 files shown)


fredrikekre:
Hmm, okay, maybe it is specific to neovim somehow then?

fredrikekre:
No probably doesn't work anywhere, I think the problem is that in the tests the client directly call the server without JSON serialization/deserialization.

davidanthoff:
There is probably no harm in adding
```julia
URI(uri::URI) = uri
```
to fix the current test failure, right?

src/requests/actions.jl
```diff
@@ -75,11 +75,11 @@ function textDocument_codeAction_request(params::CodeActionParams, server::Langu
...
 function workspace_executeCommand_request(params::ExecuteCommandParams, server::LanguageServerInstance, conn)
-    uri = params.arguments[1]
-    offset = params.arguments[2]
-    doc = getdocument(server, uri)
-    x = get_expr(getcst(doc), offset)
     if haskey(LSActions, params.command)
+        uri = URI(params.arguments[1])
+        offset = params.arguments[2]
+        doc = getdocument(server, uri)
+        x = get_expr(getcst(doc), offset)
```
pfitzseb, line 82: I've moved the uri/offset extraction into the `if` block because that "calling convention" is only valid for actions we define ourselves, and who knows what a client might send.

davidanthoff (approved):

PR #1076, opened 2022-05-05
Implement basic code action for import/using organization.

fredrikekre:
This organizes the using/import statement under the cursor, together
with any older/younger sibling expressions. The output format is close
to 'BlueStyle': import before using, and sorted alphabetically.

### DIFF
--- a/src/requests/actions.jl
+++ b/src/requests/actions.jl
@@ -54,9 +54,16 @@ function textDocument_codeAction_request(params::CodeActionParams, server::Langu
     if x isa EXPR
         for (_, sa) in LSActions
             if sa.when(x, params)
+                kind = sa.kind
+                if sa.kind !== missing && sa.kind == CodeActionKinds.SourceOrganizeImports &&
+                    server.clientInfo !== missing && occursin("code", lowercase(server.clientInfo.name))
+                    # SourceOrganizeImports doesn't show up in the VS Code UI, so make this a
+                    # RefactorRewrite instead
+                    kind = CodeActionKinds.RefactorRewrite
+                end
                 action = CodeAction(
                     sa.desc, # title
-                    sa.kind, # kind
+                    kind,    # kind
                     missing, # diagnostics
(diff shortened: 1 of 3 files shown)


pfitzseb:
Turns out that `CodeActionKinds.SourceOrganizeImports` aren't triggerable in VS Code atm (there's a setting for triggering it on save, but that's broken). Thoughts on making it a `RefactorRewrite` instead?

pfitzseb:
Turns out that `CodeActionKinds.SourceOrganizeImports` aren't triggerable in VS Code atm (there's a setting for triggering it on save, but that's broken). Thoughts on making it a `RefactorRewrite` instead?

pfitzseb:
Also needs a rebase.

fredrikekre:
It isn't triggered if you have the cursor in such a block?

pfitzseb:
Nope. CodeActions of that kind don't show up at all.

fredrikekre:
It is a bit strange that they show actions with missing kind, but not actions with "unknown" kinds. Anyway, I updated the kind if the client is VS Code, and tested that it works.

src/requests/actions.jl
```diff
@@ -54,9 +54,16 @@ function textDocument_codeAction_request(params::CodeActionParams, server::Langu
     if x isa EXPR
         for (_, sa) in LSActions
             if sa.when(x, params)
+                kind = sa.kind
+                if sa.kind !== missing && sa.kind == CodeActionKinds.SourceOrganizeImports &&
+                    server.clientInfo !== missing && server.clientInfo.name == "Visual Studio Code"
```
pfitzseb, line 59:
```suggestion
                    server.clientInfo !== missing && occursin("code", lowercase(server.clientInfo.name))
```

fredrikekre:
Yea, good catch. I wonder why that is why it doesn't show up then?

pfitzseb:
Nope, tried that.

PR #1077, opened 2022-05-06
Initial InlayHint provider implementation

pfitzseb:
image

Todo:
- [ ] better method selection
- [x] better tooltips
- [x] take request range into account
- [x] get rid of hints in func defs
- [x] settings
  - [x] none
  - [x] all
  - [x] literals only
- [x] tests


pfitzseb:
image

fredrikekre:
> settings

Perhaps it would be useful to have `constructors` as a setting? Personally I am not sure how useful hints would be for general functions, since the argument names are like an implementation detail, but when you construct something I still think it would be useful. For example: https://github.com/julia-vscode/LanguageServer.jl/blob/5b4e8e87b5fdb8afac826f4179237bca3987a0b6/src/languageserverinstance.jl#L67-L94

chriselrod:
I'm a fan of this feature in C++, looking forward to having it in Julia!

ParadaCarleton:
This would be amazing to have--any more progress?

PR #1078, opened 2022-05-06
Create initial working version of TextDocument

davidanthoff:
This is all in preparation for using `TextDocument` for different doc types.

Let's try to merge this, even though there are quite a lot of things that could still be cleaned up. But the PR is getting unwieldy enough and we can do the next steps in follow up PRs.

This PR _also_ renames the whole `offset2` story to `index`, because that is really what it is, i.e. valid Julia string indices.


pfitzseb (approved):
I like the look of this :)

PR #1080, opened 2022-05-10
Remove didSave notification handler

pfitzseb:


davidanthoff:
> I'm not sure the premise is entirely correct

What speaks in favor of the premise, though, is that we haven't seen this crash at all for months, if not years, in our crash reporting, AFAIK.

davidanthoff:
Maybe we should report this as a bug in the new client side implementation?

PR #1082, opened 2022-05-13
Actions for converting "..." strings to raw"..." strings and back.

fredrikekre:


src/requests/actions.jl
```diff
@@ -496,6 +496,76 @@ function organize_import_block(x, _, conn)
...
+            escapes -= 1
+        end
+    end
+end
+
+function convert_to_raw(x, _, conn)
+    is_string_literal(x) || return
+    file, offset = get_file_loc(x)
+    quotes = x.head === :TRIPLESTRING ? "\"\"\"" : "\"" # TODO: """ not supported yet
+    raw = "raw" * quotes * sprint(escape_raw_string, x.val) * quotes
```
pfitzseb, line 548: I personally like using `string` better than `*` and `valof(x)` instead of `x.val`, but doesn't really matter.

pfitzseb (commented):
LGTM.

Wanna add a unit test for the actual transform?

pfitzseb (approved):

PR #1093, opened 2022-05-23
Compat: Allow JuliaFormatter 0.23.X.

fredrikekre:
Doesn't seem to be anything breaking in https://github.com/domluna/JuliaFormatter.jl/compare/v0.22.11...v0.23.0, right @domluna? Should probably have been 0.22.12 then.


pfitzseb (approved):

davidanthoff:
I would just tag 1.0.0 and then one can signal much better what each release does :)

domluna:
> I would just tag 1.0.0 and then one can signal much better what each release does :)

Yeah that's a good idea

domluna:
I'm going to tag 1.0.0 after merging the SciMLStyle PR https://github.com/domluna/JuliaFormatter.jl/pull/588

davidanthoff:
Ok, after looking at things again, I think it would actually be nice if we could resolve https://github.com/domluna/JuliaFormatter.jl/issues/559 before a 1.0 tag :) Probably best to discuss that over in that issue, though.

PR #1094, opened 2022-05-23
Allow action for updating signature in docstring when cursor is in the docstring.

fredrikekre:


src/requests/actions.jl
```diff
@@ -615,9 +614,24 @@ function add_docstring_template(x, _, conn)
     return
 end
 
+function is_in_docstring_for_function(x::EXPR, _)
+    if CSTParser.isstringliteral(x) && x.parent isa EXPR && headof(x.parent) === :macrocall &&
+       length(x.parent.args) == 4 && x.parent.args[1] isa EXPR &&
+       headof(x.parent.args[1]) === :globalrefdoc && CSTParser.defines_function(x.parent.args[4])
+       return true
+    end
+    return false
```
pfitzseb, line 623:
```suggestion
    return CSTParser.isstringliteral(x) && x.parent isa EXPR && headof(x.parent) === :macrocall &&
       length(x.parent.args) == 4 && x.parent.args[1] isa EXPR &&
       headof(x.parent.args[1]) === :globalrefdoc && CSTParser.defines_function(x.parent.args[4])
```

pfitzseb (approved):

PR #1104, opened 2022-06-17
Use loose_refs in for_each_ref

pfitzseb:
This is generally closer to what people expect when talking about references.
Before:
image

After:
image


davidanthoff:
Is the idea that we no longer follow scoping rules? I always assume that this command gives me the same list I would get from the rename refactoring, which presumably follows scoping rules? So not sure I’m a fan of this change :) Do we have evidence that the current implementation is confusing a lot of users?

pfitzseb:
This change also affects refactoring.

The issue with the current implementation is that we *don't* correctly follow Julia's scoping rules. You'd end up going from 
```
function ff()
    x = 0
    while true
        y = sin(x)
        x = y
        while true
            y = sin(x)
            x = y
        end
        x = 2y
    end
end
```
to e.g. 
```
function ff()
    x = 0
    while true
        yyy = sin(x)
        x = yyy
        while true
            y = sin(x)
            x = y
        end
        x = 2y
    end
end
```
which changes the meaning of the code -- the inner loop doesn't update the outer loops `y` (now `yyy`) after all. With this change you'd get 
``` 
function ff()
    x = 0
    while true
        yyy = sin(x)
        x = yyy
        while true
            yyy = sin(x)
            x = yyy
        end
        x = 2yyy
    end
end
```
instead. Basically, the difference between `.refs` and `loose_refs` is that the latter is more correct for soft scopes.

PR #1105, opened 2022-06-17
Re-enable incremental sync and sync checking

pfitzseb:


davidanthoff (approved):
Should we merge and push an insider build to see how it goes?

PR #1107, opened 2022-06-18
Make URI case insensitive for files on win

### LINKED ISSUE #1103: No diagnostics / new URI handling maybe faulty for Windows drive letters?
### Background

I just updated LanguageServer.jl to the latest commit 512b05add26a5d83fc2aca74a8c820c65385fd09 for the Sublime Text client (Manifest.toml temporarily [here](https://github.com/sublimelsp/LSP-julia/blob/test/server/Manifest.toml)).

### Symptoms

After the language server initializes, diagnostics are reported as expected. But as soon as you open a Julia file, the language server reports empty diagnostics for that file.

### Investigation

Just a speculation, but perhaps this may be caused by an updated URI handling in the language server (https://github.com/julia-vscode/LanguageServer.jl/commit/7b4759b7c20235ae1f98fc46de380e35534d963c)? The following seems suspicious to me:

From the JSON-RPC log I can see that the language server usually reports diagnostics with lowercase drive letter, for example `file:///d%3A/ ...`
```
:: <-  julia textDocument/publishDiagnostics: {'version': 0, 'diagnostics': [{'message': 'An argument is included in a function signature but not used within its body.', 'severity': 4, 'code': 'UnusedFunctionArgument', 'range': {'end': {'line': 0, 'character': 117}, 'start': {'line': 0, 'character': 113}}, 'source': 'Julia', 'tags': [1]}], 'uri': 'file:///d%3A/git/LanguageServer.jl/src/requests/highlight.jl'}
```

But Sublime Text uses an uppercase drive letter for the URIs, for example
```
::  -> julia textDocument/didOpen: {'textDocument': {'languageId': 'julia', 'version': 0, 'uri': 'file:///D:/git/LanguageServer.jl/src/requests/highlight.jl', 'text': 'function textDocument_documentHighlight_request(params::DocumentHighlightParams, server::LanguageServerInstance, conn)\n    doc = getdocument(server, params.textDocument.uri)\n    offset = get_offset(doc, params.position)\n    identifier = get_identifier(getcst(doc), offset)\n    identifier !== nothing || return nothing\n    highlights = DocumentHighlight[]\n    for_each_ref(identifier) do ref, doc1, o\n        if get_uri(doc1) == get_uri(doc)\n            kind = StaticLint.hasbinding(ref) ? DocumentHighlightKinds.Write : DocumentHighlightKinds.Read\n            push!(highlights, DocumentHighlight(Range(doc, o .+ (0:ref.span)), kind))\n        end\n    end\n    return isempty(highlights) ? nothing : highlights\nend\n'}}
```

After that, the language server will send back a notification with empty diagnostics for that file. Note the uppercase drive letter now, similar to what was just sent by the client:
```
:: <-  julia textDocument/publishDiagnostics: {'version': 0, 'diagnostics': [], 'uri': 'file:///D%3A/git/LanguageServer.jl/src/requests/highlight.jl'}
```

Interestingly, other requests seem to still work fine, e.g. documentHighlight, code actions etc., even though the client always uses the uppercase drive letter in its requests.

davidanthoff:
Fixes #1103.

I'm not super happy about this implementation (also never was about the previous one), as it feels off to have this code as part of the `URI` type. But for now, good enough.

### DIFF
--- a/src/URIs2/URIs2.jl
+++ b/src/URIs2/URIs2.jl
@@ -12,6 +12,36 @@ struct URI
     fragment::Union{String,Nothing}
 end
 
+@static if Sys.iswindows()
+    function Base.:(==)(a::URI, b::URI)
+        if a.scheme=="file" && b.scheme=="file"
+            a_path_norm = lowercase(a.path)
+            b_path_norm = lowercase(b.path)
+
+            return a.scheme == b.scheme &&
+                a.authority == b.authority &&
+                a_path_norm == b_path_norm &&
+                a.query == b.query &&
+                a.fragment == b.fragment
+        else
+            return a.scheme == b.scheme &&
+                a.authority == b.authority &&
+                a.path == b.path &&
+                a.query == b.query &&
+                a.fragment == b.fragment
+        end
+    end
+
+    function Base.hash(a::URI, h::UInt)
+        if a.scheme=="file"
+            path_norm = lowercase(a.path)
+            return hash((a.scheme, a.authority, path_norm, a.query, a.fragment), h)
+        else
+            return hash((a.scheme, a.authority, a.path, a.query, a.fragment), h)
+        end
+    end
+end
+
 function percent_decode(str::AbstractString)
     return URIs.unescapeuri(str)
 end


jwortmann:
Thanks for the fix! I've tested it and can confirm that it fixes the linked issue.

pfitzseb (approved):

PR #1145, opened 2022-08-19
Add more checks for NULL in strings

davidanthoff:
Should fix a crash reporting bug, plus I added the same check in all places where we call `isvalid` as that seems what we want.


pfitzseb:
Probably a good idea to factor that out into its own function.

davidanthoff:
Yep, good point. Now in a separate function.

PR #1172, opened 2022-10-15
Add some diagnostic code for signature help

davidanthoff:
Trying to get to the bottom of a crash report with this.


davidanthoff:
@pfitzseb I'm going to merge now because I want to get an insider build out, but would be great if you would review after the fact.

PR #1176, opened 2022-10-16
Use unreserved error codes

davidanthoff:
`-32099` is actually in the reserved range of error codes, and for example is already used by the ls client library that we use in the extension on the Typescript side of things for a different error condition, so one can no longer differentiate between these error conditions on the client side.

This now uses an error code that should be outside any reserved range. Technically we'll need to tag a major version, unless we are very positive that no client other than VS Code is using this functionality?

Also needs a corresponding PR on the extension side.

### DIFF
--- a/src/utilities.jl
+++ b/src/utilities.jl
@@ -3,7 +3,7 @@
 
 function nodocument_error(uri, data=nothing)
     return JSONRPC.JSONRPCError(
-        1000,
+        -33100,
         "document $(uri) requested but not present in the JLS",
         data
     )
@@ -11,7 +11,7 @@ end
 
 function mismatched_version_error(uri, doc, params, msg, data=nothing)
     return JSONRPC.JSONRPCError(
-        1001,
+        -33101,
         "version mismatch in $(msg) request for $(uri): JLS $(get_version(doc)), client: $(params.version)",
         data
     )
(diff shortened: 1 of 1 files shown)


src/utilities.jl
```diff
@@ -1,12 +1,17 @@
...
+        1000,
+        "document $(uri) requested but not present in the JLS",
+        data
+    )
+end
 
 function mismatched_version_error(uri, doc, params, msg, data=nothing)
     return JSONRPC.JSONRPCError(
-        -32099,
+        1001,
```
pfitzseb, line 14: Thoughts on keeping in line with error codes in the -32k (ish) range? E.g. `-32101` etc?

davidanthoff: I’m reading the spec as saying those are all reserved for future use and therefore shouldn’t be used by us?

pfitzseb: Ah, right, skipped over that paragraph. -33xxx then?

davidanthoff: I have zero preferences, so also happy to use that range. I picked the ones in this PR because they seemed far away from anything anyone else might use. Are there conversations for negative numbers around, in general?

davidanthoff: Alright, now with new error codes :)

pfitzseb (approved):

Issue #1180, opened 2022-10-20
Range formatting doesn't work with (0,0) as the initial position


mroavi:
I'm using Neovim 0.8 together with Julia's LSP server (LanguageServer 4.3.1)
I've noticed a few issues when trying to format code using `vim.lsp.buf.format` and its `range` argument.

```
format({options})                                       *vim.lsp.buf.format()*
    Formats a buffer using the attached (and optionally filtered) language
    server clients.

    Parameters:  
        {options}  table|nil Optional table which holds the following optional
                   fields:
                   .....
            
                   • range (table|nil) Range to format. Table must contain
                     `start` and `end` keys with {row, col} tuples using (1,0)
                     indexing. Defaults to current selection in visual mode
                     Defaults to `nil` in other modes, formatting the full
                     buffer 
```

Here is the first of a series:

Before:
```julia
function clr()
for _ = 1:60
println()
end
end
```
After `:lua vim.lsp.buf.format({ range = { ['start'] = {1,0}, ['end'] = {5,0} } })`
```julia
function clr()
for _ = 1:60
println()
end
end
```

Expected:
```julia
function clr()
  for _ = 1:60
    println()
  end
end
```

fredrikekre:
For me it works if you add an empty line before. It looks like it doesn't work when the initial position in the range is `(0, 0)`.

mroavi:
The issue is that `function get_inner_expr` is returning an empty expression.

While I don't understand the code, I did figure out why it returns an empty expression:

```julia
# full (not only trivia) expr containing rng, modulo whitespace
function get_inner_expr(x, rng::UnitRange{Int}, pos=0, pos_span = 0)
    if all(pos .> rng)
        # DOES NOT ENTER HERE
        return nothing
    end
    if length(x) > 0 && headof(x) !== :NONSTDIDENTIFIER
        # ENTERS HERE!
        pos_span′ = pos_span
        for a in x
            if a in x.args && all(pos_span′ .< rng .<= (pos + a.fullspan))
                # DOES NOT ENTER HERE
                return get_inner_expr(a, rng, pos, pos_span′)
            end
            pos += a.fullspan
            pos_span′ = pos - (a.fullspan - a.span)
        end
    elseif pos == 0
        return x
    elseif all(pos_span .< rng .<= (pos + x.fullspan))
        return x
    end
    pos -= x.fullspan
    if all(pos_span .< rng .<= (pos + x.fullspan))
[...]

Issue #1188, opened 2022-10-28
Incorrect range formatting in file with do-block


mroavi:
Related to https://github.com/julia-vscode/LanguageServer.jl/issues/1180 and https://github.com/julia-vscode/LanguageServer.jl/issues/1187

Before:
```julia
open("out", "w") do io
write(io, data)
end

function hello()
println("Hello, World")
end
```

After `:lua vim.lsp.buf.format({ range = { ['start'] = {1,0}, ['end'] = {3,0} } })`
```julia
open("out", "w") do io
  write(io, data)
end

function hello()
  println("Hello, World")
end
```

Expected:
```julia
open("out", "w") do io
  write(io, data)
end

function hello()
println("Hello, World")
end
```

mroavi:
https://github.com/julia-vscode/LanguageServer.jl/pull/1185/commits/b905089676d1f497f9344a2d4160633c45932360 fixes this.

mroavi:
I'll continue posting problems with range formatting in this issue instead of opening new ones.

Here is another case that causes the entire file to be formatted:

`:lua vim.lsp.buf.format({ range = { ['start'] = {1,0}, ['end'] = {3,0} } })`

```julia
# open("out", "w") do io
# write(io, data)
# end

function hello()
println("Hello, World")
end
```

mroavi:
Another more worrying case:

Before:
```julia
function printx(x)

            println("x = $x")
            return nothing

end
```

After `:lua vim.lsp.buf.format({ range = { ['start'] = {3,0}, ['end'] = {4,0} } })`
```julia
function printx(x)

            print
            return nothing

end
```

Issue #1187, opened 2022-10-28
Entire file gets formatted when the range refers to the top most text-block of code


mroavi:
I'm using Neovim 0.8 together with Julia's LSP server (LanguageServer 4.3.1)

Before:
```julia
 
function hi()
println("Hi, World")
end

function hello()
println("Hello, World")
end
```

After: `:lua vim.lsp.buf.format({ range = { ['start'] = {2,0}, ['end'] = {4,0} } })`
```julia
 
function hi()
  println("Hi, World")
end

function hello()
  println("Hello, World")
end
```

Expected:
```julia
 
function hi()
  println("Hi, World")
end

function hello()
println("Hello, World")
end
```

fredrikekre:
I am pretty sure the `do` is a red herring and this is just a version of #1184?

mroavi:
Like in https://github.com/julia-vscode/LanguageServer.jl/issues/1180, the issue is inside the `get_inner_expr` function. In this case, it returns an expression of the entire file rather than of the range argument.

mroavi:
I just noticed that the do-block is not what is causing the issue.

mroavi:
> I am pretty sure the `do` is a red herring and this is just a version of #1184?

Very probable. @danielsoutar does your issue occur only when selecting the top-most block of code? In my case, if the given range encloses the `hello` function, then the formatting is done correctly.

mroavi:
The PR above fixes this problem in the same way that @pfitzseb suggested in https://github.com/julia-vscode/LanguageServer.jl/issues/1180 but in a different if statement.

danielsoutar:
> > I am pretty sure the `do` is a red herring and this is just a version of #1184?
> 
> Very probable. @danielsoutar does your issue occur only when selecting the top-most block of code? This is what I'm seeing. For example, if the range I pass includes only the `hello` function, then the formatting is done correctly.

I use VSCode, and for my workflow I highlight the selection, and choose 'Format selection' via the command bar. My issue was when I tried formatting across several functions. This is with a custom .JuliaFormatter.toml file, so I'd need to check with my work machine whether this applies with the default formatter.

danielsoutar:
Apologies for cluttering this thread, but can I get an answer on how I can bring this change into my VSCode extension? Could I check out a particular version of the language server to get this change, or do I need to wait for an official release?

mroavi:
[...]

Issue #1184, opened 2022-10-27
Format selection just formats the entire document


danielsoutar:

fredrikekre:
I think if you select a top-level expression (for example a function) then the entire top level scope is formatted. But if you select e.g. a loop inside a function it only formats the selection.

PR #1190, opened 2022-10-31
Fix range formatting

mroavi:
Fixes https://github.com/julia-vscode/LanguageServer.jl/issues/1180, https://github.com/julia-vscode/LanguageServer.jl/issues/1187, https://github.com/julia-vscode/LanguageServer.jl/issues/1188, and probably https://github.com/julia-vscode/LanguageServer.jl/issues/1184

Here is the idea:
1. We take as input the entire file and the range that should be formatted.
2. We enclose the given range by injecting two markers in the form of comments to the input file.
3. We format the result of step 2 (entire file with markers) using JuliaFormatter.jl. The output should include the injected markers.
4. We replace the given range in the input file with the code inside the markers of the output of step 3, and return this.

This approach avoids the need of having to deal with EXPRs and formats the given range in the context of the entire file.

### DIFF
--- a/src/requests/features.jl
+++ b/src/requests/features.jl
@@ -214,7 +214,7 @@ function textDocument_range_formatting_request(params::DocumentRangeFormattingPa
     if isnothing(range_formatted)
         return TextEdit[]
     end
-    return TextEdit[TextEdit(Range(params.range.start.line, 0, params.range.stop.line, params.range.stop.character), range_formatted[1])]
+    return TextEdit[TextEdit(Range(params.range.start.line, 0, params.range.stop.line, 99999999), range_formatted[1])]
 end
 
 function find_references(textDocument::TextDocumentIdentifier, position::Position, server)
--- a/src/requests/features.jl
+++ b/src/requests/features.jl
@@ -184,10 +184,10 @@ function format_text(text::AbstractString, params, config)
 end
 
 function mark_range(text::AbstractString, startline, stopline, startmark, stopmark)
-  textlines = split(text, "\n")
-  insert!(textlines, startline + 1, startmark)
-  insert!(textlines, stopline + 3, stopmark)
-  return join(textlines, "\n")
+    textlines = split(text, "\n")
+    insert!(textlines, startline + 1, startmark)
+    insert!(textlines, stopline + 3, stopmark)
+    return join(textlines, "\n")
 end
 
 function textDocument_range_formatting_request(params::DocumentRangeFormattingParams, server::LanguageServerInstance, conn)
@@ -197,8 +197,8 @@ function textDocument_range_formatting_request(params::DocumentRangeFormattingPa
     oldcontent = get_text(doc)
     startline = params.range.start.line
     stopline = params.range.stop.line
-    startmark =  "#___________START___________"
-    stopmark =   "#___________STOP____________"
+    startmark = "#___________START___________"
+    stopmark = "#___________STOP____________"
     text_marked = mark_range(oldcontent, startline, stopline, startmark, stopmark)
 
     text_formatted = try
@@ -214,9 +214,9 @@ function textDocument_range_formatting_request(params::DocumentRangeFormattingPa
 
     range_formatted = match(Regex("$startmark\n((?s).*\n)\\s*$stopmark"), text_formatted)
     if isnothing(range_formatted)
-      newcontent = oldcontent
+        newcontent = oldcontent
     else
-      newcontent = replace(text_marked, Regex("$startmark\n(?s).*$stopmark\n") => range_formatted[1])
+        newcontent = replace(text_marked, Regex("$startmark\n(?s).*$stopmark\n") => range_formatted[1])
     end
     end_l, end_c = get_position_from_offset(doc, sizeof(get_text(doc))) # AUDIT: OK
     lsedits = TextEdit[TextEdit(Range(0, 0, end_l, end_c), newcontent)]
--- a/src/requests/features.jl
+++ b/src/requests/features.jl
@@ -195,8 +195,8 @@ function textDocument_range_formatting_request(params::DocumentRangeFormattingPa
     oldcontent = get_text(doc)
     startline = params.range.start.line + 1
     stopline = params.range.stop.line + 1
-    startmark = "#؞؞؞؞؞؞؞؞؞LANGUAGESERVER_RANGE_FORMATTING_START؞؞؞؞؞؞؞؞؞"
-    stopmark = "#؞؞؞؞؞؞؞؞؞LANGUAGESERVER_RANGE_FORMATTING_STOP؞؞؞؞؞؞؞؞؞"
+    startmark = "#" * string(uuid4())
+    stopmark = "#" * string(uuid4())
     text_marked = mark_range(oldcontent, startline, stopline, startmark, stopmark)
 
     text_formatted = try
--- a/src/requests/features.jl
+++ b/src/requests/features.jl
@@ -185,9 +185,10 @@ end
 
 function mark_range(text::AbstractString, startline, stopline, startmark, stopmark)
     textlines = split(text, "\n")
+    stopline_size = sizeof(textlines[stopline])
     insert!(textlines, stopline + 1, stopmark)
     insert!(textlines, startline, startmark)
-    return join(textlines, "\n")
+    return join(textlines, "\n"), stopline_size
 end
 
 function textDocument_range_formatting_request(params::DocumentRangeFormattingParams, server::LanguageServerInstance, conn)
@@ -197,7 +198,7 @@ function textDocument_range_formatting_request(params::DocumentRangeFormattingPa
     stopline = params.range.stop.line + 1
     startmark = "#" * string(uuid4())
     stopmark = "#" * string(uuid4())
-    text_marked = mark_range(oldcontent, startline, stopline, startmark, stopmark)
+    text_marked, stopline_size = mark_range(oldcontent, startline, stopline, startmark, stopmark)
 
     text_formatted = try
         config = get_juliaformatter_config(doc, server)
@@ -216,7 +217,7 @@ function textDocument_range_formatting_request(params::DocumentRangeFormattingPa
     if isnothing(range_formatted) || (range_formatted[1] == range_unformatted[1])
         return TextEdit[]
     end
-    return TextEdit[TextEdit(Range(params.range.start.line, 0, params.range.stop.line, 99999999), range_formatted[1])]
+    return TextEdit[TextEdit(Range(params.range.start.line, 0, params.range.stop.line, stopline_size), range_formatted[1])]
 end
 
 function find_references(textDocument::TextDocumentIdentifier, position::Position, server)
--- a/src/requests/features.jl
+++ b/src/requests/features.jl
@@ -191,9 +191,7 @@ function mark_range(text::AbstractString, startline, stopline, startmark, stopma
 end
 
 function textDocument_range_formatting_request(params::DocumentRangeFormattingParams, server::LanguageServerInstance, conn)
-
     doc = getdocument(server, params.textDocument.uri)
-
     oldcontent = get_text(doc)
     startline = params.range.start.line
     stopline = params.range.stop.line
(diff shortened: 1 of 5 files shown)


src/requests/features.jl
```diff
@@ -183,51 +183,25 @@ function format_text(text::AbstractString, params, config)
     end
 end
 
+function mark_range(text::AbstractString, startline, stopline, startmark, stopmark)
+    textlines = split(text, "\n")
+    insert!(textlines, startline + 1, startmark)
+    insert!(textlines, stopline + 3, stopmark)
```
pfitzseb, line 189:
```suggestion
    insert!(textlines, stopline + 1, stopmark)
    insert!(textlines, startline, startmark)
```

src/requests/features.jl
```diff
@@ -183,51 +183,25 @@ function format_text(text::AbstractString, params, config)
...
-            end
-        end
-        line = line[1:idx]
-        longest_prefix = CSTParser.longest_common_prefix(something(longest_prefix, line), line)
-    end
-
-    newcontent = try
+    oldcontent = get_text(doc)
+    startline = params.range.start.line
+    stopline = params.range.stop.line
```
pfitzseb, line 197:
```suggestion
    startline = params.range.start.line + 1
    stopline = params.range.stop.line + 1
```

src/requests/features.jl
```diff
@@ -183,51 +183,25 @@ function format_text(text::AbstractString, params, config)
...
+    startmark = "#___________START___________"
+    stopmark = "#___________STOP____________"
```
pfitzseb, line 199: Not sure how relevant it is, but we could use UUIDs here to reduce the chance of a collision with genuine comments. Or scan the text for these first I guess.

src/requests/features.jl
```diff
@@ -236,16 +210,14 @@ function textDocument_range_formatting_request(params::DocumentRangeFormattingPa
...
+        newcontent = oldcontent
+    else
+        newcontent = replace(text_marked, Regex("$startmark\n(?s).*$stopmark\n") => range_formatted[1])
     end
-
-    lsedits = TextEdit[TextEdit(Range(l1, c1, l2, c2), newcontent)]
-
+    end_l, end_c = get_position_from_offset(doc, sizeof(get_text(doc))) # AUDIT: OK
+    lsedits = TextEdit[TextEdit(Range(0, 0, end_l, end_c), newcontent)]
     return lsedits
```
pfitzseb, line 221: I think this can just be 
```
    range_formatted = match(Regex("$startmark\n((?s).*)\n\\s*$stopmark"), text_formatted)
    if isnothing(range_formatted)
        return TextEdit[]
    end
    return TextEdit[TextEdit(Range(params.range.start.line, 0, params.range.stop.line, 99999999), range_formatted[1])]
```
so we don't replace the whole text document. `99999999` should probably be the actual line length though :)

pfitzseb (commented):
Generally LGTM and works very nicely. I've left a few comments inline.

src/requests/features.jl
fredrikekre: Perhaps just use something more specific like
```suggestion
    startmark = "#___________LANGUAGESERVER_RANGE_FORMATTING_START___________"
    stopmark = "#___________LANGUAGESERVER_RANGE_FORMATTING_STOP____________"
```

src/requests/features.jl
mroavi: Ahh yes. This is better. By the way, I was thinking that perhaps it would be nice to avoid editing the file if the formatting makes no change to the text in the range. This would avoid marking the file as dirty when no change was made.

src/requests/features.jl
mroavi: I ended up using the UUID approach. Let me know what you think.

src/requests/features.jl
```diff
@@ -236,17 +210,13 @@ function textDocument_range_formatting_request(params::DocumentRangeFormattingPa
...
+    range_formatted = match(range_regex, text_formatted)
+    range_unformatted = match(range_regex, text_marked)
+    if isnothing(range_formatted) || (range_formatted[1] == range_unformatted[1])
+        return TextEdit[]
     end
-
-    lsedits = TextEdit[TextEdit(Range(l1, c1, l2, c2), newcontent)]
-
-    return lsedits
+    return TextEdit[TextEdit(Range(params.range.start.line, 0, params.range.stop.line, 99999999), range_formatted[1])]
```
pfitzseb, line 219: You can get the end char as a return value from `mark_range` (`sizeof(textlines[stopline])`)

mroavi: `mark_range` does not have access to the formatted range.
What about getting it from the `range_formatted` text like this: `sizeof(last(split(range_formatted[1], "\n")))`

pfitzseb: We don't care about the formatted text. This is a range into the unformatted text that's replaced.

mroavi: or `sizeof(range_unformatted[1]) - findlast('\n', range_unformatted[1])`

mroavi: Ah, got it.

mroavi:
This last commit introduced a bug that the tests does not catch.
```julia
function textDocument_range_formatting_request(params::DocumentRangeFormattingParams, server::LanguageServerInstance, conn)
    doc = getdocument(server, params.textDocument.uri)
    oldcontent = get_text(doc)
    startline = params.range.start.line + 1
    stopline = params.range.stop.line + 1
    startmark = "#" * string(uuid4())
    stopmark = "#" * string(uuid4())
    text_marked, stopline_size  = mark_range(oldcontent, startline, stopline, startmark, stopmark)
```
is formatted to
```julia
function textDocument_range_formatting_request(params::DocumentRangeFormattingParams, server::LanguageServerInstance, conn)
    doc = getdocument(server, params.textDocument.uri)
    oldcontent = get_text(doc)
    startline = params.range.start.line + 1
    stopline = params.range.stop.line + 1
    startmark = "#" * string(uuid4())
    stopmark = "#" * string(uuid4())
    text_marked, stopline_size = mark_range(oldcontent, startline, stopline, startmark, stopmark)rk_range(oldcontent, startline, stopline, startmark, stopmark)
```

src/requests/features.jl
mroavi: Ahhh got it.. it should be `sizeof(textlines[stopline + 1])`

mroavi:
After the last changes, this case stopped working:

`:lua vim.lsp.buf.format({ range = { ['start'] = {1,0}, ['end'] = {6,0} } })`

Input:
```julia
function printx(x)

println("x = $x")
return

end
```

Output:
```julia
function printx(x)

println("x = $x")
return

endend
```

Since now we are only replacing the text inside the given range, it is more difficult to test because we cannot specify how the entire file has to look after formatting (see https://github.com/julia-vscode/LanguageServer.jl/pull/1190/commits/c05eaa7d794889143d002278355f41fd599971bf). This case is not caught by any of the tests because the returned formatted text is correct. However, it is misplaced in the input file. Any idea how we could make these tests more robust? I will look into why this case fails later this week.

fredrikekre:
> the returned formatted text is correct

Where does the extra `end` come from then?

fredrikekre:
What do you mean with copied over? Sounds like a mismatch in how you mark what is going to be formatted then? The `end` is included in the text to be formatted, but excluded from the range of replacement?

mroavi:
> What do you mean with copied over? Sounds like a mismatch in how you mark what is going to be formatted then? The `end` is included in the text to be formatted, but excluded from the range of replacement?

I mean that the `end` should be overwritten with the formatted `end`, which in this case looks exactly the same. I'm not sure where the issue comes from. It could be indeed an issue with how the range is marked. It could also be that the marking is being done properly but not the way in which we are replacing the text within the marks with the formatted text. I'll take a closer look today when I get home.

fredrikekre:
I mean that since you insert regular line comments you must mark full lines to format, and then you must also replace the full line, not the original range, right?

mroavi:
> I mean that since you insert regular line comments you must mark full lines to format

Yes.

> and then you must also replace the full line

we need to replace the introduced marks (comments) and all the lines inside the range.

We use this regex to do that: 

https://github.com/mroavi/LanguageServer.jl/blob/1a12fc86426d8d37bcb554a8f2244ed32486793d/src/requests/features.jl#L214-L216

I see a `\n` in there which might be the issue. I think it should be optional in order to cover this case.

mroavi:
And this is the other part that we modified:

https://github.com/mroavi/LanguageServer.jl/blob/1a12fc86426d8d37bcb554a8f2244ed32486793d/src/requests/features.jl#L220

which could also be the source of the problem. I don't have the laptop configured to debug LS.jl at the moment. I'll post an update as soon as I can test this a bit more.

mroavi:
I'm sorry. This was all noise. The range formatting is working fine. I think I accidentally typed the undo button and disregarded the fix made here https://github.com/julia-vscode/LanguageServer.jl/commit/1a12fc86426d8d37bcb554a8f2244ed32486793d. 😳

fredrikekre:
Yea I couldn't reproduce. I played around a bit with the PR, is it OK if I push some changes to hopefully make it more resilient towards different lineendings?

mroavi:
Please do so!

src/requests/features.jl
```diff
@@ -183,23 +183,26 @@ function format_text(text::AbstractString, params, config)
...
 end
 
-function mark_range(text::AbstractString, startline, stopline, startmark, stopmark)
-    textlines = split(text, "\n")
-    stopline_size = sizeof(textlines[stopline])
-    insert!(textlines, stopline + 1, stopmark)
-    insert!(textlines, startline, startmark)
-    return join(textlines, "\n"), stopline_size
-end
+# Broken up to make this file formattable
```
mroavi, line 186: Not entirely sure what you mean by this.

fredrikekre: The two constant strings below are broken up and joined with `*` so that the marker doesn't exist in this file :) I will clarify the comment.

fredrikekre (approved):

HusterRichard:
When there is an empty line at the end of the code, the format selection operation will be performed. There will be the following error:
ERROR: BoundsError: attempt to access 3-element Vector{String} at index [1:4]
Stacktrace:
 [1] throw_boundserror(A::Vector{String}, I::Tuple{UnitRange{Int64}})
   @ Base .\abstractarray.jl:691
 [2] checkbounds
   @ .\abstractarray.jl:656 [inlined]
 [3] view
   @ .\subarray.jl:177 [inlined]
 [4] textDocument_range_formatting_request(params::LanguageServer.DocumentRangeFormattingParams, server::LanguageServerInstance, conn::JSONRPC.JSONRPCEndpoint{Base.PipeEndpoint, Base.PipeEndpoint})
   @ LanguageServer c:\Users\admin\.vscode\extensions\julialang.language-julia-1.47.2\scripts\packages\LanguageServer\src\requests\features.jl:198
 [5] (::LanguageServer.var"#111#112"{typeof(LanguageServer.textDocument_range_formatting_request), LanguageServerInstance})(conn::JSONRPC.JSONRPCEndpoint{Base.PipeEndpoint, Base.PipeEndpoint}, params::LanguageServer.DocumentRangeFormattingParams)
   @ LanguageServer c:\Users\admin\.vscode\extensions\julialang.language-julia-1.47.2\scripts\packages\LanguageServer\src\languageserverinstance.jl:267
 [6] dispatch_msg(x::JSONRPC.JSONRPCEndpoint{Base.PipeEndpoint, Base.PipeEndpoint}, dispatcher::JSONRPC.MsgDispatcher, msg::Dict{String, Any})
   @ JSONRPC c:\Users\admin\.vscode\extensions\julialang.language-julia-1.47.2\scripts\packages\JSONRPC\src\typed.jl:67
 [7] run(server::LanguageServerInstance)
   @ LanguageServer c:\Users\admin\.vscode\extensions\julialang.language-julia-1.47.2\scripts\packages\LanguageServer\src\languageserverinstance.jl:387
 [8] top-level scope
   @ c:\Users\admin\.vscode\extensions\julialang.language-julia-1.47.2\scripts\languageserver\main.jl:104

fredrikekre:
Fixed by #1228?

PR #1210, opened 2023-03-01
dot methods completion  

xgdgsc:
Fixes https://github.com/julia-vscode/LanguageServer.jl/issues/979  .

 Updates https://github.com/julia-vscode/LanguageServer.jl/pull/980.  
completes 
```
dframea = DataFrame()
dframea.
```
show list, then select `append!` to 
```
append!(dframea,)
```

A few issues remain:

1. cannot search the list
2. the change in line 16 needed to show the list, I don' t know if it's right to change this.
3.  whether it' s ok to use dot as trigger.
4.  Some types don' t  work, like Vector
5.  move cursor to middle of ",<CURSOR>)"
6.  Would it be better to call methodswith in REPL mode?

I remember some language that does this in IDE ( cannot remember which one for now) . I think it' s a convenient way for people coming from OOP to find the relevant methods. I' m looking for suggestions and opinions.

### DIFF
--- a/src/requests/completions.jl
+++ b/src/requests/completions.jl
@@ -6,17 +6,18 @@ struct CompletionState
     offset::Int
     completions::Dict{String,CompletionItem}
     range::Range
-    x::Union{Nothing, EXPR}
+    x::Union{Nothing,EXPR}
     doc::Document
     server::LanguageServerInstance
     using_stmts::Dict{String,Any}
 end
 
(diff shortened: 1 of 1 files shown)


xgdgsc:
Seems a common CI error. It works with latest https://github.com/julia-vscode/julia-vscode/pull/3240 for me.

PR #1218, opened 2023-04-26
Add PrecompileTools, precompile runserver

mkitti:
Add PrecompileTools to and precompile `runserver`.


mkitti:
I'm not sure why Project.toml got rearranged, but it should only be adding PrecompileTools.jl

mkitti:
```
julia> using PkgCacheInspector, Pkg

julia> Pkg.develop("LanguageServer")

julia> info_cachefile("LanguageServer")
Contents of /home/mkitti/.julia/compiled/v1.9/LanguageServer/ite7n_4MdHE.so:
  modules: Any[LanguageServer.URIs2, LanguageServer]
  815 external methods
  3756 new specializations of external methods (Base 72.7%, StaticLint 10.1%, JSONRPC 5.0%, ...)
  666 external methods with new roots
  15443 external targets
  11613 edges
  file size:   17084984 (16.294 MiB)
  Segment sizes (bytes):
  system:      5087440 ( 51.21%)
  isbits:      4201270 ( 42.29%)
  symbols:       59097 (  0.59%)
  tags:          73137 (  0.74%)
  relocations:  454313 (  4.57%)
  gvars:         30664 (  0.31%)
  fptrs:         28448 (  0.29%)
```

timholy:
Nice. The small amount of residual compilation described in https://discourse.julialang.org/t/precompiling-languageserver-runserver-with-precompiletools-jl/97926 might be fixable by `@nospecializing(pipe)` on all the IO objects in LanguageServer. (Presumably it doesn't actually run with a `IOBuffer`.) If you can `@snoopi_deep` it under "real" usage you can more directly figure out what wasn't covered by your workload.

pfitzseb:
Very cool! Did you time the difference in startup time?

And yes, we initialize the LS with `Base.TTY`s instead of `IOBuffer`s.

timholy:
If printing doesn't have to be high performance, `@nospecializing` could be a good solution. Otherwise, there's been a request already to set expand PrecompileTools to add a harness capable of instantiating a fake TTY. If there are enough potential customers, adding that harness becomes more worth the downside of making PrecompileTools bigger.

timholy:
> Very cool! Did you time the difference in startup time?

In https://discourse.julialang.org/t/precompiling-languageserver-runserver-with-precompiletools-jl/97926, he shows that `@time runserver()` goes from 30s to 10s. `using LanguageServer` goes from 20ms to 500ms, so it seems like a clear win.

davidanthoff (approved):
Fantastic! I'd say we can merge right away. I'll add PrecompileTools to the extension later today with our submodule way of loading things.

Also, we can probably improve _a lot_ on this, because just running the server probably only executes a relatively small fraction of the code. But let's do that in follow up PRs.

fredrikekre:
Can 
```
┌ LanguageServer [2b0e0bc5-e4fd-59b4-8912-456d1b03d8d7]
│  ERROR: Endpoint is not running, the current state is closed.
│  Stacktrace:
│   [1] error(s::String)
│     @ Base ./error.jl:35
│   [2] check_dead_endpoint!
│     @ ~/.julia/packages/JSONRPC/Q0FBr/src/core.jl:280 [inlined]
│   [3] get_next_message(endpoint::JSONRPC.JSONRPCEndpoint{IOBuffer, IOContext{IOStream}})
│     @ JSONRPC ~/.julia/packages/JSONRPC/Q0FBr/src/core.jl:212
│   [4] macro expansion
│     @ ~/.julia/packages/LanguageServer/bnyQN/src/languageserverinstance.jl:288 [inlined]
│   [5] (::LanguageServer.var"#113#116"{LanguageServer.LanguageServerInstance})()
│     @ LanguageServer ./task.jl:514
│  [ Info: Shutting down server instance.
└
```
be silenced? (I assume it comes from this PR)

mkitti:
I'm missing the part of the error stack trace that connects it with this pull request. Does this happen during precompilation?

fredrikekre:
Yes

mkitti:
One option from the code below is to set an error handler. Otherwise, we'll need to modify the `else` block below.

https://github.com/julia-vscode/LanguageServer.jl/blob/198e07ed89d49360f187c315157c5582411904a8/src/languageserverinstance.jl#L293-L299

mkitti:
Is it just me or does the language server _always_ end with an error?

https://github.com/julia-vscode/LanguageServer.jl/blob/198e07ed89d49360f187c315157c5582411904a8/src/languageserverinstance.jl#L287-L290

PR #1253, opened 2023-10-22
Add Kate editor to list

fbanning:
This PR adds Kate to the list of supported editors. The wiki has already been updated with an entry about how to use it and what to prepare.

NB: The PR ([here](https://invent.kde.org/utilities/kate/-/merge_requests/1324)) has just been merged recently, so it will take a bit of time until most users' Kate installations will have the default settings available to them. In the meantime the setting can be easily added by hand (about which I will add a note to the wiki as well).


fbanning:
P.s.: I've tweaked the Wiki entry a bit ([see here](https://github.com/julia-vscode/LanguageServer.jl/wiki/Kate)) and hopefully this should be enough info to get people started with Kate's LSP config.

pfitzseb (approved):

PR #1268, opened 2024-02-19
Add better support for SPDX headers

Seelengrab:
This commit does these things:

 * Add rudimentary detection of EUPL-* licenses
 * Add logging statements throughout the action to make it debuggable should some license not be properly detected
 * Fix a bug with SPDX header detection in existing files

To expand on the last point - previously, the header detection didn't have multiline matching enabled on its regex, which meant that any file having any lines other than the SPDX header wasn't detected as actually having a header. That's a bit bad, since presumably there would be actual source code following that header.

For a visual example, before:

image

After:

image

As for tests - I looked at `test/requests/test_actions.jl` and found an extremely minimal test for that action. I have no idea how to add a test for the functionality there, so I'd appreciate any guidance you can offer.

### DIFF
--- a/src/requests/actions.jl
+++ b/src/requests/actions.jl
@@ -350,7 +350,9 @@ function double_to_triple_equal(x, _, conn)
 end
 
 function get_spdx_header(doc::Document)
-    m = match(r"(*ANYCRLF)^# SPDX-License-Identifier:\h+((?:[\w\.-]+)(?:\h+[\w\.-]+)*)\h*$", get_text(doc))
+    # note the multiline flag - without that, we'd try to match the end of the _document_
+    # instead of the end of the line.
+    m = match(r"(*ANYCRLF)^# SPDX-License-Identifier:\h+((?:[\w\.-]+)(?:\h+[\w\.-]+)*)\h*$"m, get_text(doc))
     return m === nothing ? m : String(m[1])
 end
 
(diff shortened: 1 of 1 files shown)


Seelengrab:
Since this bumps compat to 1.3 for the artifact, the failure on <1.3 is entirely expected - I hope this can still go through in some form, either actually dropping support for 1.0 or in some manner that allows better detection of licenses other than MIT.

pfitzseb:
Unfortunately we're fairly committed to keeping 1.0 compat for now.

So, while being able to identify the license is probably useful, I think just fixing the regex is probably best for now. Also cc @fredrikekre because I think he originally implemented this?

Seelengrab:
That's unfortunate to hear, especially since 1.0 has been out of language support for 2+ years now. I'd have loved to do some PRs with other code actions, but I don't really want to go back to the time of slow compilation :)

I'll adjust the PR so only a very minimal version is done. Specifically:

 * Fixing the regex, so the SPDX header is identified correctly
 * Adding detection of EUPL licenses the way it's done for MIT

pfitzseb (approved):

pfitzseb:
Nope! Thanks again for the fix.

fredrikekre:
> I'd have loved to do some PRs with other code actions

I think it is fine to have some actions behind a version check so don't let the Julia 1.0 support discourage you.

pfitzseb:
Yes, but that's a bit tricky when it comes to dependencies.

PR #1270, opened 2024-02-25
Refactor Error Code definitions

Seelengrab:
This refactors the way error codes are handled, in preparation for making JSONRPC.jl less LSP specific, by hooking into custom error printing provided by an upstream PR. Requires https://github.com/julia-vscode/JSONRPC.jl/pull/70.

### DIFF
--- a/src/LanguageServer.jl
+++ b/src/LanguageServer.jl
@@ -17,6 +17,69 @@ using .URIs2
 
 JSON.lower(uri::URI) = string(uri)
 
+# these are the endpoints of JSON-RPC error codes
+# per the LSP spec, no LSP error codes should lie
+# between these
+const SERVER_ERROR_END       = -32000
+const SERVER_ERROR_START     = -32099
+
+# These are existing reserved errors for JSONRPC.jl
+# We shouldn't throw these manually
+const SERVER_NOT_INITIALIZED = -32002
+const UNKNOWN_ERROR_CODE     = -32001
+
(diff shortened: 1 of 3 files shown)


Seelengrab:
There's also these error codes that seem to overlap considerably:

https://github.com/julia-vscode/LanguageServer.jl/blob/e67befb27eb0783a7dcbef11eca8513d15e3a0a5/src/requests/features.jl#L155-L167

This error code is seemingly used for a formatting failure, but the [LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#errorCodes) says this should not be used:

```
/**
 * This is the end range of JSON-RPC reserved error codes.
 * It doesn't denote a real error code.
 *
 * @since 3.16.0
 */
export const jsonrpcReservedErrorRangeEnd = -32000;
```

This should probably be changed to some other error code, but I don't know how that would interact with the extension.

Similarly, this section also defines a new error code for formatting:

https://github.com/julia-vscode/LanguageServer.jl/blob/e67befb27eb0783a7dcbef11eca8513d15e3a0a5/src/requests/features.jl#L207-L217

This doesn't conflict with the spec, but it doesn't match the existing formatting error code used above. This needs a decision on whether they should be merged into one error code, or kept distinct.

pfitzseb:
Both of those should be `-33000` I think. As far as I'm aware no-one is actually checking for the specific error code anyways, FWIW.

PR #1271, opened 2024-03-09
Add check for `params.capabilities.window.workDoneProgress`

VarLad:
Adds a minor check for `params.capabilities.window.workDoneProgress`. 
This was done because some editors only partially support this optional protocol.


pfitzseb (approved):

pfitzseb:
Thanks!

PR #1273, opened 2024-03-18
Organize imports: put module self-binding first

fredrikekre:
This patch changes the organize import action to put any self-bindings first in the sorted import list. Before this patch sorting results in

```julia
using Example: AbstractThing, Example
```

and after the patch

```julia
using Example: Example, AbstractThing
```


pfitzseb (approved):
LGTM, but should maybe be a function?

fredrikekre:
Fixed.

pfitzseb (approved):

PR #1313, opened 2024-08-16
Fix server doesn't initialize

jwortmann:
Closes #1310.

Disclaimer: only tested on Windows.

PR #1319, opened 2024-08-27
Extract project files tracking into function

jwortmann:
I noticed that the drive letter normalization from my PR #1313 was incomplete and also needs to be added for the environment changing via `julia/activateenvironment` here:
https://github.com/julia-vscode/LanguageServer.jl/blob/0da181a8e48da14c2ac46519311e01352c7c6fca/src/requests/misc.jl#L85-L109

This exact same code is used in two places, so I extracted it into a new function `track_project_files!` (the diff for `misc.jl` on GitHub is unfortunately a bit misleading).


jwortmann:
I split the PR up into 2 commits, to make it easier to review. I'd recommend to squash the commits when merging.

jwortmann:
It would be great if this PR could be merged, because this bug is currently blocking the update of the server for clients that use uppercase drive letters.

jwortmann:
Bump :shrug:

Does anybody read issues and pull requests in this repository and is able to merge this?
The changes are really simple and it shouldn't take more than two minutes or so to review.

PR #1324, opened 2024-09-30
Show warnings when a request is slow

DanielVandH:
As discussed @davidanthoff , this outputs a warning when the duration of a request exceeds 1 second. I put the warning here instead of `request_wapper` so that the request name has a better formatting.


DanielVandH:
Don't really know the best way to add a test for this.. I've tested it locally by just waiting for a slow request but is there something better to do to simulate a slow request in the tests?

PR #1327, opened 2024-10-23
[WIP] add unicode completions by latex name

tqml:
# Feature: Unicode variable completion by latex name (WIP)

This PR adds the ability to work with variables that use unicode symbols as names more conveniently. As an example consider there exists a variable `θ₀`. With this PR, the symbol `θ₀` would be added to code completion suggestions if the user types a string close to `theta_0` (which is easier on keyboards that dont feature greek characters) 

Demo:

https://github.com/user-attachments/assets/ad230bb4-69b4-4f6c-a1ad-112fb9e0fec7


tqml:
Change log updated here:

julia-vscode/julia-vscode#3720

jwortmann:
I don't take any decisions over which PRs should or shouldn't be merged in this repository, but are you aware that the language server already provides the unicode symbols when you type the LaTeX name (or an emoji name), but starting with a backslash, i.e. like in actual LaTeX or in the Julia REPL?

https://github.com/julia-vscode/LanguageServer.jl/blob/0da181a8e48da14c2ac46519311e01352c7c6fca/src/requests/completions.jl#L53-L56

Also I think it is quite common to use the greek letter names written in ASCII as variable names (like `alpha`, `beta`, `mu`), and then I imagine that it would be problematic if they are automatically converted to unicode symbols when auto-completing.

tqml:
> I don't take any decisions over which PRs should or shouldn't be merged in this repository, but are you aware that the language server already provides the unicode symbols when you type the LaTeX name (or an emoji name), but starting with a backslash, i.e. like in actual LaTeX or in the Julia REPL?
> 
> https://github.com/julia-vscode/LanguageServer.jl/blob/0da181a8e48da14c2ac46519311e01352c7c6fca/src/requests/completions.jl#L53-L56
> 
> Also I think it is quite common to use the greek letter names written in ASCII as variable names (like `alpha`, `beta`, `mu`), and then I imagine that it would be problematic if they are automatically converted to unicode symbols when auto-completing.

Thanks for the comment!

Yes, I'm aware about it but its not what this PR is about (sorry, I think my description is not clear enough in this regard)

This PR actually tries to do something different: The completion __only__ occurs, when an existing symbol __already exists__, so just typing `alpha` would not autocomplete to the unicode symbol.  So in my example above, `alpha` would not autocomplete while `zeta` and `omega` do, because there are already unicode symbols defined, that match that name.

The very first symbol needs to be typed by latex-backslash way.  However, having to use the latex-backslash __all of the time__ slows you down quite a bit (in my experience), thats why I usually stop using unicode symbols all together and just use ASCII after a while (which is a shame, because I think its a really nice feature!)
I also admit, that this is mostly due to my keyboard (German, Mac), where the backslash is just at an inconvenient position.

I'm open for comments and feedback because this is a very opinionated change (maybe it would make sense to have a preference toggle for it).

jwortmann:
Okay I understand now, thanks for the explanation. Then I guess it could be something useful. Maybe it should do another check to not provide this unicode conversion if there is also an existing symbol with the corresponding ASCII name already defined. Though that would probably be a quite bad and confusing code style if names like e.g. `θ` and `theta` are defined at the same time, so I don't know how realistic such a use case is.

tqml:
Good point, that could confusing. 
Not sure what the right thing to do there is, I tend to solution to just show both symbols but maybe with some ranking/order, e.g. for the query `theta` would match a variable named `theta` exactly and should get a higher score/rank than `θ` (because here it matches only the alternative name)

Actually, I think some kind of ranking/suggestion for the completion suggestion would make sense overall but is not implemented from what I saw.

Issue #1316, opened 2024-08-20
Wrong `Manifest.toml` file used in Julia 1.11


frankebel:
Hi, I'm using the language server in the editor [Neovim](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#julials). So far, everything workes fine on Julia 1.10.

For testing, I also installed Julia 1.11-rc2 on the side which supports [version specific Manifest files](https://github.com/JuliaLang/julia/blob/v1.11.0-rc2/NEWS.md) which I use. Thus, I have 2 Manifest files inside the root of the project:
- `Manifest.toml` for Julia 1.10
- `Manifest-v1.11.toml` for Juila 1.11

The REPL can differentiate the files as advertised, but `LanguageServer.jl` always picks up `Manifest.toml` even if I start it with Julia 1.11.

PR #1372, opened 2026-01-17
Support version-specific Manifest

quachpas:
Fixes https://github.com/julia-vscode/LanguageServer.jl/issues/1316. This is in parallel with https://github.com/julia-vscode/SymbolServer.jl/pull/303. Same comments apply.

> This adds to the list of manifest candidates the [version-specific Manifest](https://pkgdocs.julialang.org/v1/toml-files/#:~:text=Different%20Manifests%20for%20Different%20Julia%20versions) based on the Julia version of the LSP.
> There is a strong assumption because in load_project_packages_into_store!, SymbolServer.jl will now prefer in order: Manifest-vX.XX.toml, Manifest.toml, JuliaManifest.toml.
> I am not sure whether this is acceptable. If this is approved, then I will documentation at the VS Code extension docs and mention this in the changelog.

### DIFF
--- a/src/requests/init.jl
+++ b/src/requests/init.jl
@@ -211,7 +211,7 @@ function initialized_notification(params::InitializedParams, server::LanguageSer
             Registration("workspace/didChangeWatchedFiles", "workspace/didChangeWatchedFiles", DidChangeWatchedFilesRegistrationOptions([
                 FileSystemWatcher("**/*.{jl,jmd,md}", missing),
                 FileSystemWatcher("**/{Project.toml,JuliaProject.toml,Manifest.toml,JuliaManifest.toml,.JuliaLint.toml}", missing),
-                FileSystemWatcher("**/Manifest-v$(VERSION.major).$(VERSION.minor).toml", missing),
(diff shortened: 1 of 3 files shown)


src/requests/init.jl
```diff
@@ -210,7 +210,8 @@ function initialized_notification(params::InitializedParams, server::LanguageSer
             client_capabilities_registrations,
             Registration("workspace/didChangeWatchedFiles", "workspace/didChangeWatchedFiles", DidChangeWatchedFilesRegistrationOptions([
                 FileSystemWatcher("**/*.{jl,jmd,md}", missing),
-                FileSystemWatcher("**/{Project.toml,JuliaProject.toml,Manifest.toml,JuliaManifest.toml,.JuliaLint.toml}", missing)
+                FileSystemWatcher("**/{Project.toml,JuliaProject.toml,Manifest.toml,JuliaManifest.toml,.JuliaLint.toml}", missing),
+                FileSystemWatcher("**/Manifest-v$(VERSION.major).$(VERSION.minor).toml", missing),
```
pfitzseb: I don't think we want to separate watchers here. Should also probably be watching all `JuliaManifest-vx.y`, but I'm not sure if `**/{JuliaManifest,Manifest}-v{[0-9],[0-9].[0-9]}.toml` is going to work.

PR #1393, opened 2026-05-29
fix: handle requests against documents not present in the sesrver

pfitzseb:


davidanthoff:
Is this a good idea? It means we no longer get a crash report, and unless the client has a bug, this presumably only happens if there is some root problem in our code?

pfitzseb:
Best I can tell this is a client bug, yes. I'd be happy to still send telemetry, but crashing seems problematic here imo.

davidanthoff:
In what situation did this occur? A non VS Code client?

pfitzseb:
Yes, when opening a notebook in Theia.

PR #1418, opened 2026-08-10
Disable the special casing of the Julia folder

davidanthoff:
Apparently this is outdated in any case. Also, I want to get to a point where the JW is actually usable on the Julia repo. I have Claude on it ;)


pfitzseb (approved):
No idea what this is about...

PR #1467, opened 2026-09-11
Only offer formatting for Julia documents

davidanthoff:
## Crash signature

From the julia-vscode **insider-channel** Application Insights crash telemetry, last 30 days. This is
the single largest crash signature in the entire corpus:

| Signature | Events | Users |
| --- | --- | --- |
| `Error at handleResponse` — `Failed to format document: …` | **3645** | 19 |

That is ~70% of all reported extension "crashes" for the period. A representative message:

```
Failed to format document: ErrorException("ParseError:
# Error @ line 3:… constant horizontal body force drives the flow throughout the …
#└────────────────────────────────────────┘ ── extra tokens after end of expression")
```

The text being "parsed" is English prose: these are **Markdown files** being sent to the Julia
formatter, overwhelmingly by format-on-save.

## Root cause

The client's document selector covers `julia`, `juliamarkdown` **and** `markdown` (all three get
diagnostics, completions, and so on). `ServerCapabilities` advertised formatting statically:

```julia
documentFormattingProvider      = true
documentRangeFormattingProvider = true
```

A static capability applies to the client's *whole* document selector, so the server registered as a
formatting provider for plain Markdown as well. Two consequences:

1. `get_format_edits` parses the entire document as Julia, so formatting a Markdown file always fails
   and returns a `JSONRPCError`, which the client rethrows and VS Code reports as an extension fault.
2. More importantly, an editor picks **one** formatting provider per document. By claiming the
   capability we were also *suppressing the real Markdown formatter*. Simply returning "no edits" would
   not have fixed that — we have to not be registered for those languages at all.

## Fix

**Scope the registration.** Advertise `documentFormattingProvider` / `documentRangeFormattingProvider`
as `missing` whenever the client supports dynamic registration, and instead register
`textDocument/formatting` and `textDocument/rangeFormatting` from `initialized_notification` with a
document selector limited to `language: julia`. The `client/registerCapability` machinery was already
there for `workspace/didChangeWatchedFiles`.

Formatting a Markdown file now falls through to whichever formatter the user actually configured.

`juliamarkdown` (`.jmd`) is excluded too: it is Markdown with Julia chunks, so formatting it as a single
Julia file fails the same way. It keeps every other language-server feature.

Clients that do **not** advertise dynamic registration keep the previous static capability rather than
silently losing formatting. The two decisions are made by the same predicate so they cannot drift apart,
and the capabilities are decided independently of each other.

**Report syntax errors as user-facing failures.** A smaller slice of these events are genuine `.jl`
files that do not parse. That is the user's code, not a server bug, but `-32000` with a stringified
Julia exception is the wrong way to say so. These now return LSP 3.17 `RequestFailed` (`-32803`) with a
message written for a human and naming the file.

Note this alone does not stop the crash report: `vscode-languageclient`'s `handleFailedRequest` only
swallows `PendingResponseRejected`, `ConnectionInactive`, `RequestCancelled`, `ServerCancelled` and
`ContentModified`, and logs+rethrows everything else including `RequestFailed`. The client-side half is
a companion PR against julia-vscode that catches this in formatting middleware and shows the user a
warning.

## Testing

`test/requests/test_formatting.jl`:

- Updated the existing syntax-error test to assert the new `RequestFailed` code, that the message names
  the file, and that it no longer leaks the old internal wording.
- Added `formatting is only offered for Julia documents`, covering: the selector contains `julia` and
  neither `markdown` nor `juliamarkdown`; a dynamic-registration client gets `missing` for both
  capabilities; `missing` is **omitted from the serialised JSON** rather than sent as `null` (otherwise
  the client would still see a provider); a non-dynamic client and a client that advertises nothing both
  keep `true`; the two capabilities are decided independently; and the emitted registration serialises to
  the expected LSP wire shape with the Julia-only selector.

```
LanguageServer                                                                         |   27     27
  test/requests/test_formatting.jl                                                     |   27     27
    formatting an excluded file is a no-op, not an error                               |    4      4
    formatting a file with a syntax error reports RequestFailed for both request kinds |    8      8
    formatting is only offered for Julia documents                                     |   15     15
```

### DIFF
--- a/src/requests/features.jl
+++ b/src/requests/features.jl
@@ -14,17 +14,33 @@ function textDocument_definition_request(params::TextDocumentPositionParams, ser
     return unique!(locations)
 end
 
+"""
+LSP 3.17 `ErrorCodes.RequestFailed`: the request was well-formed and understood,
+but could not be carried out. The message is meant to be shown to the user, so it
+must read as an explanation rather than as an internal exception dump.
+"""
+const LSP_REQUEST_FAILED = -32803
+
+# A formatting failure is almost always "this file is not valid Julia", which is
+# the user's code, not a server fault. Lead with the actionable sentence and keep
+# the underlying parser output after it for the logs.
+function format_failure_error(uri::URI, err)
+    name = something(uri2filepath(uri), string(uri))
+    detail = sprint(showerror, err)
+    return JSONRPC.JSONRPCError(
+        LSP_REQUEST_FAILED,
+        "Could not format $(basename(name)) because it could not be parsed as Julia code.\n\n$detail",
+        nothing
+    )
+end
+
 function textDocument_formatting_request(params::DocumentFormattingParams, server::LanguageServerInstance, conn)
     uri = params.textDocument.uri
 
     file_edit = try
         JuliaWorkspaces.get_format_edits(server.workspace, uri)
     catch err
-        return JSONRPC.JSONRPCError(
-            -32000,
-            "Failed to format document: $err.",
-            nothing
-        )
+        return format_failure_error(uri, err)
     end
 
     # A file the configuration excludes is not an error; the gesture simply
@@ -45,11 +61,7 @@ function textDocument_range_formatting_request(params::DocumentRangeFormattingPa
     file_edit = try
         JuliaWorkspaces.get_format_edits(server.workspace, uri, start_line, stop_line)
     catch err
-        return JSONRPC.JSONRPCError(
-            -32000,
-            "Failed to format document: $err.",
-            nothing
-        )
+        return format_failure_error(uri, err)
     end
 
     file_edit === nothing && return TextEdit[]
--- a/src/requests/init.jl
+++ b/src/requests/init.jl
@@ -201,7 +201,10 @@ function initialized_notification(params::InitializedParams, server::LanguageSer
     # from using the formatter that actually can (it picks a single provider per
     # document). `ServerCapabilities` leaves the static capability out whenever the
     # client advertises dynamic registration, so the two must stay in sync.
-    if !ismissing(server.clientCapabilities)
+    #
+    # `conn === nothing` means there is no client to register with (the server is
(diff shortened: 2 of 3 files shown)


davidanthoff:
### Note for maintainers: a latent test-harness conflict this PR ran into

CI initially failed on all 24 platforms here, and the cause is worth recording because it is
pre-existing and will catch the next change that touches `initialized_notification`.

`JSONRPC.send(::Nothing, ...)` is defined **18 times across 8 test files**, all with the same
signature:

- `test_shared_server.jl` (`TestSetup`), `test_indirect_files.jl` x3, `test_watched_folders.jl` x4 define it as a no-op
- `test_indexing_complete.jl`, `test_initialized_publish.jl` x3, `test_progress.jl` x3 define it as `push!(sent, (typ, params))`, capturing into a per-testitem `sent` vector

Because these are method definitions on a shared function and the items run in the same process, they
overwrite each other globally and the last one loaded wins. When a capturing definition is active for a
test item that has no `sent` of its own, any send fails with:

```
MethodError: no method matching push!(::Nothing, ::Tuple{JSONRPC.RequestType{RegistrationParams, Nothing}, RegistrationParams})
```

This never fired before because nothing in `initialized_notification` actually sent anything under the
`SharedServer` setup: that client advertises neither `didChangeConfiguration` nor
`didChangeWatchedFiles` dynamic registration, so the registration batch was always empty. Adding the
formatting registration made it non-empty for the first time and immediately tripped the conflict, in
`TextDocument`, `editor pid monitoring (#1379)`, `publish sweep: diffs against published state` and
one of the new formatting tests.

I have sidestepped it correctly for this PR (no client connection means there is nothing to register
with), rather than papering over it. But the underlying harness issue remains: a single shared capture
buffer in `TestSetup` -- e.g. `const SENT = Any[]` with tests calling `empty!(TestSetup.SENT)` --
would remove the method piracy entirely. That felt out of scope for a crash fix, so I am flagging it
rather than bundling it.

